PR #2, opened 2023-07-28
Support older versions of PHP

jessarcher:


jessarcher:
Note to self: we'd need to replace `array_is_list` usage to support PHP 8.0, and we'd need to remove union types to support anything lower.

PR #5, opened 2023-07-28
Added support for PHP 8

msh1596:
Added support to use Laravel Prompt on PHP 8 environment.


msh1596:
Supporting PHP 7.3 and 7.4, means that we can not use anymore named arguments.

taylorotwell:
Prompts is PHP 8.1+ for now.

PR #9, opened 2023-07-28
(feat) Add `P` class with static methods

lucasmichot:
This pull-request add a P class, that can be imported to call any function.
It is aligned with [laravel/framework](https://github.com/laravel/framework) where all the string and array helpers functions were migrated to dedicated `Arr` and `Str` classes with static methods.

That change also allows to greatly limit the number of imports in a console command class, where https://github.com/laravel/prompts feature would be used.

**Note:**
The playground and tests files have also been updated to reflect this change.

### DIFF
--- a/playground/confirm.php
+++ b/playground/confirm.php
@@ -1,10 +1,10 @@
 <?php
 
-use function Laravel\Prompts\confirm;
+use Laravel\Prompts\P;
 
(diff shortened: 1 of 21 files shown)


taylorotwell:
Not sure I love this idea.

PR #12, opened 2023-07-28
(feat) No need to set up and GA strategy without a matrix.

lucasmichot:
This pull-request removes the useless GitHub Action strategy.


taylorotwell:
I think it's OK.

PR #14, opened 2023-07-28
Replace tput command as it's not available on all OSes

digiservnet:
The `tput` command isn't available by default in the Alpine Linux official PHP docker images (at least) and this throws an `InvalidArgumentException` for truncation length.

This PR replaces the `tput` command with method calls from the Symfony Console `Terminal` class.

`tput` is used in the `Terminal` class to determine the number of rows and columns available, in particular used for truncating line length if available terminal columns are too low.

### DIFF
--- a/src/Terminal.php
+++ b/src/Terminal.php
@@ -2,6 +2,8 @@
 
 namespace Laravel\Prompts;
 
+use Symfony\Component\Console\Terminal as Term;
+
 class Terminal
 {
     /**
@@ -62,7 +64,7 @@ class Terminal
      */
     public function cols(): int
     {
-        return $this->cols ??= (int) shell_exec('tput cols 2>/dev/null');
+        return $this->cols ??= (new Term())->getWidth();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jessarcher:
Thanks!

I wonder if we could change the approach or add a fallback...

For example, `stty size` returns the lines and columns separated by a space on Linux, and we're already using `stty` elsewhere. Does that work in Alpine?

I'd just need a macOS friend to check whether that works and returns the same format for them, and then we could just drop `tput` altogether.

digiservnet:
Tested and `stty size` works fine in Alpine. If this can be confirmed on a mac (from looking around on the interwebz, it does produce the same `rows cols` format), I can push the changes. 🙂

digiservnet:
Had confirmation that `stty` behaves the same on macOS so have gone with this approach.

Also added Windows detection and utilised `mode` there. The output from this is:
```bash
mode CON
```

```bash

Status for device CON:
----------------------
    Lines:          42
    Columns:        180
    Keyboard rate:  31
    Keyboard delay: 0
    Code page:      850
```

digiservnet:
Good call @wouterj. I've replaced the custom code with the Symfony Terminal `getWidth()` and `getHeight()` method calls as there doesn't seem much reason to duplicate this functionality 🙂

jessarcher:
Thanks!

Issue #13, opened 2023-07-28
Can't type "0" on a text input: Uninitialized string offset 0


igorblumberg:
### Laravel Prompts Version

0.1.0

### Laravel Version

10.16.1

### PHP Version

8.2.6 

### Operating System & Version

Ubuntu (Laravel Homestead)

### Terminal Application

Oh My Zsh

### Description

When I prompt the user to enter a text, if the character "0" (zero) is present in the input the command fails immediately. 

$userID = text(
            label: "Type the user ID"
        );
$this->info($userID);

### Steps To Reproduce

Run the command and type "0" in the input field.

   ErrorException 

  Uninitialized string offset 0

  at vendor/laravel/prompts/src/Concerns/TypedValue.php:43
     39▕             $this->cursorPosition = strlen($this->typedValue);
     40▕         }
     41▕ 
     42▕         $this->on('key', function ($key) use ($submit) {
  ➜  43▕             if ($key[0] === "\e") {
     44▕                 match ($key) {
     45▕                     Key::LEFT => $this->cursorPosition = max(0, $this->cursorPosition - 1),
     46▕                     Key::RIGHT => $this->cursorPosition = min(strlen($this->typedValue), $this->cursorPosition + 1),
     47▕                     Key::DELETE => $this->typedValue = substr($this->typedValue, 0, $this->cursorPosition).substr($this->typedValue, $this->cursorPosition + 1),

      +5 vendor frames 
  6   app/Console/Commands/Commit.php:30
      Laravel\Prompts\text()

      +12 vendor frames 
  19  artisan:35
      Illuminate\Foundation\Console\Kernel::handle()

PR #15, opened 2023-07-29
fix: Check for non-null return value from fread to handle character z…

igorblumberg:
As detailed here #13 the input fails if "0" is typed. 

This is happening because fread(STDIN, 1024) is returning "0" but fread(STDIN, 1024) ?:"" returns "".

### DIFF
--- a/src/Terminal.php
+++ b/src/Terminal.php
@@ -24,8 +24,9 @@ class Terminal
      */
     public function read(): string
     {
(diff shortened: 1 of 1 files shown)


jessarcher:
Thanks! :pray: 

I solved a similar bug in e979362 but then regressed it in ffe3ae6 :woman_facepalming:

PR #30, opened 2023-08-02
Support Laravel 9

danharrin:
Many packages in the Laravel ecosystem still support Laravel 9 (Livewire, Inertia, Filament, many Spatie packages with commands), and it would be great if they could make use of prompts to enhance their UX.

I have ensured that GitHub actions runs tests using Laravel 9, and they appear to be passing. Please let me know if I've made an oversight here and there is a reason why it's not currently supported.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - main
       - '*.x'
-      - l9
   pull_request:
   schedule:
     - cron: '0 0 * * *'
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,8 +20,9 @@ jobs:
       fail-fast: true
       matrix:
         php: [8.1, 8.2]
+        laravel: [9, 10]
 
-    name: PHP ${{ matrix.php }}
+    name: PHP ${{ matrix.php }} - Laravel ${{ matrix.laravel }}
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hmm, I think my main hesitation is this binds us to Laravel 9 until the next major release of Prompts - not letting us take advantage of anything that is introduced to Laravel 10 in the meantime.

danharrin:
That's fair enough, but the only new features you would actually be able to take advantage of are from `illuminate/collection`, and everything that has already been achieved with prompts so far doesn't require L10.to

My issue is that to be able to use prompts, we have to drop Laravel 9 support from packages, while its still a security-supported release by you guys, and I don't think users would be too happy with that. So in the end, we just won't end up using prompts, which would be a shame.

At the end of the day having this is not a huge deal, but it would be really nice if we could make good use of all the fantastic work that has been done with this package, and push it out to loads more users.

awcodes:
Couldn’t this have possible impacts on Livewire too if it still supports Laravel 9?

jhonoryza:
@danharrin can you add support for php 8.0 i am still using laravel 9 with php 8.0 :)

HassanZahirnia:
Please approve this PR, wanna use this in our projects as well 🙌 Thank you 😊

taylorotwell:
With Laravel 11 coming this winter I don't want to bind us to Laravel 9. If the upgrade path from 9 to 10 was significant I would agree with this but the upgrade path is relatively trivial and takes just a few minutes.

PR #33, opened 2023-08-03
support mutli-byte characters

### LINKED ISSUE #29: Multi-byte characters make the right border position incorrect
### Laravel Prompts Version

v0.1.1 (I also tested dev-main)

### Laravel Version

Independent installation of prompts

### PHP Version

8.2.8

macOS 13.0

### Terminal Application

Terminal

### Description

When I input multi-byte characters (such as Chinese, Japanese, Emoji, etc.), the width of the border at the end changes. These multibyte characters occupy 2 spaces, but the calculation boundary may use its actual length (3 spaces).

Maybe using `mb_strwidth` to solve this?

### Steps To Reproduce

image

image

tyler36:
This PR adds support for multi-btye characters:
- emojis
- Japanese languages (Hiragana, Katakana, Kanji)
- Chinese languages

This fixes #29 

## Example

The below code produces the following screenshot.

```php
use function Laravel\Prompts\confirm;

$confirmed = confirm(
  label: '楽しんでいますか？',
  yes: 'はい',
  no: 'いいえ',
);
var_dump($confirmed);
```

prompt-before

## After PR

Appling the PR produces the following:

prompt-after

I'm not sure if/how we can automate testing this but the current tests are still green.

### DIFF
--- a/src/Concerns/Truncation.php
+++ b/src/Concerns/Truncation.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Laravel\Prompts\Concerns;
+
+use InvalidArgumentException;
+
+trait Truncation
+{
+    /**
+     * Truncate a value with an ellipsis if it exceeds the given width.
+     */
+    protected function truncate(string $string, int $width): string
+    {
+        if ($width <= 0) {
+            throw new InvalidArgumentException("Width [{$width}] must be greater than zero.");
+        }
+
+        return mb_strwidth($string) <= $width ? $string : (mb_strimwidth($string, 0, $width - 1).'…');
+    }
+}
--- a/src/Concerns/TypedValue.php
+++ b/src/Concerns/TypedValue.php
@@ -85,30 +85,38 @@ trait TypedValue
      */
     protected function addCursor(string $value, int $cursorPosition, int $maxWidth): string
     {
-        $offset = $cursorPosition - $maxWidth + ($cursorPosition < mb_strlen($value) ? 2 : 1);
-        $offset = $offset > 0 ? $offset + 1 : 0;
-        $offsetCursorPosition = $cursorPosition - $offset;
+        $before = mb_substr($value, 0, $cursorPosition);
+        $current = mb_substr($value, $cursorPosition, 1);
+        $after = mb_substr($value, $cursorPosition + 1);
 
-        $output = $offset > 0 ? $this->dim('…') : '';
-        $output .= mb_substr($value, $offset, $offsetCursorPosition);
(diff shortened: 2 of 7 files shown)


jessarcher (changes requested):
Thanks!

Just tested on the text prompt, and the cursor position can get a bit broken. It seems to disappear at the end.

I think for anything to do with cursor position, we'd still want to use `mb_strlen` because an emoji, etc., should only count as one character in that scenario. I think it's just the padding where they should count as 2 characters.

tyler36:
Thanks for the update. 

I revert the change for the "prompt" classes. So `mb_strwidth` is only applied to the theming files: ‘Renderer.php`, and `DrawsBoxes.php`.

Does that resolve the problem for you?

jessarcher:
The cursor issue seems fixed, but now it breaks when the characters exceed the terminal's width. I think it's a bit more complex than I'd hoped. We might need to go back to using `mb_strwidth` in more places, but then also use `mb_strimwidth` instead of `mb_substr`.

crazywhalecc:
I have previously written a program in PHP for terminal output of multi-byte text, it may need some special detection and calculation. In short, we may need a structure to mark the text input length, text display length and text data length. 😵‍💫 This is just my thoughts, I have not read all the prompts code.

taylorotwell:
@jessarcher just mark this as ready for review when ready.

jessarcher:
Hi @tyler36 and @crazywhalecc,

Can you please test the latest commit if you get a chance?

I ended up rewriting the virtual cursor and horizontal text scrolling logic, but it should be a lot easier to understand the next time I need to look at it.

jessarcher (approved):

tyler36:
@jessarcher 

I ran through a quick test with 0c30f08f933521021243d93094326e717014482e

## Test: text

```
$name = text('楽しんでいますか？');
var_dump($name );
```

The box correctly grew and shrank with a mixutre of Japanese (hiragana and kanji) and English. 
`$name` correctly dumped.

prompt-multibyte

Cursor behavior felt "normal" with extremely long Japanese lorem:
`$name` correctly dumped.
prompt-ja-lorem

## Test: confirm

Confirm box also displayed correctly.

```php
$confirmed = confirm(
  label: '楽しんでいますか？',
  yes: 'はい',
  no: 'いいえ',
);
var_dump($confirmed);
```

See screenshot under screenshot above for comparison with extended "text" prompt.
`$confirmed` correctly dumped.

tyler36:
Looks good to me. Thank you!

taylorotwell:
Thanks!

PR #36, opened 2023-08-03
Add info note

sebastianpopp:
I'm using prompts outside of Laravel and think it would be helpful to be able to display a simple success message the same way I can display an error message. The name of the function is based on the function in Laravel commands.

### DIFF
--- a/playground/notes.php
+++ b/playground/notes.php
@@ -2,6 +2,7 @@
 
 use function Laravel\Prompts\alert;
 use function Laravel\Prompts\error;
+use function Laravel\Prompts\info;
 use function Laravel\Prompts\intro;
 use function Laravel\Prompts\note;
 use function Laravel\Prompts\outro;
@@ -11,6 +12,7 @@ require __DIR__.'/../vendor/autoload.php';
 
 intro('Intro');
 note('Note');
+info('Info');
 warning('Warning');
 error('Error');
 alert('Alert');
--- a/src/Themes/Default/NoteRenderer.php
+++ b/src/Themes/Default/NoteRenderer.php
@@ -42,6 +42,11 @@ class NoteRenderer extends Renderer
 
                 return $this;
 
+            case 'info':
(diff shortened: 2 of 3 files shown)


jessarcher:
Thanks for the PR, @sebastianpopp!

It's cool to hear you're using Prompts in other PHP projects and that you found the note components!

I'm okay with merging this, but I'm not 100% sure what the future holds for the notes (and spinner), which are currently undocumented. I think they're very useful, but they don't completely replace the informational components from Symfony, like tables and progress bars. I haven't had time to think about where the "non-prompting" parts of Prompts fit and the value they can bring. In any case, I don't anticipate us removing them, but they could potentially take a different shape in a later release.

jessarcher (approved):

utsavsomaiya:
In the Laravel application, we have a mix and match situation concerning helper function names. Currently, the helper function info is used to interact with the Log channel. To enhance clarity and avoid conflicts, I suggest renaming this helper function to success. By doing so, we can easily differentiate between logging information and indicating a successful operation.

what do you think @jessarcher?

jessarcher:
> In the Laravel application, we have a mix and match situation concerning helper function names. Currently, the helper function `info` is used to interact with the Log channel. To enhance clarity and avoid conflicts, I suggest renaming this helper function to `success`. By doing so, we can easily differentiate between logging information and indicating a successful operation.
> 
> what do you think @jessarcher?

I don't have a strong opinion. The prompt functions are all namespaced, so there is no conflict. The `info` name matches the corresponding `$this->info()` method in Artisan console commands, but I understand it's not ideal to share a name as Laravel's global `info` function and need to alias an import if you want to use both in the same file.

I'll need to think more about the role of these functions and have a chat with Taylor. I probably should have left them out of the package until I had a clearer picture for them.

utsavsomaiya:
No worries dear. Thanks.

Issue #42, opened 2023-08-07
Any luck with more natural scroll list?


crazywhalecc:
I yesterday tried to contribute to the project to support a natural scroll method, and I realized there is no view state variable. For detail:

1. My expectation: Only when I scroll to an item outside of the displayed one, do I scroll to the displayed item on the page.
2. Actual: When I scroll the list (select, multiselect, search), and options count is bigger than height, the highlighted option is always last one of view, related code: https://github.com/laravel/prompts/blob/a7fc438f5a1be415a669493b151805daee2d25bd/src/Themes/Default/Concerns/DrawsScrollbars.php#L47
3. And I tried to add a variable to render, I realize renderer creates every input time. So I also added a static instance variable to persist this object, but another bug appeared: it will top up the previous window.
4. As far as I know, the Prompt class is the only one that can be used as storing view state.

I would be happy if you could provide some ideas or if this issue could be resolved appropriately. I don't know if this is more reasonable.

jessarcher:
This is something I had hoped to achieve, but I was already finding it quite challenging to position the scrollbar just based on the number of items and the currently selected item, ensuring that the scrollbar is never in the first or last position until you reach the first or last item. The added state required to do what you're talking about, and all the possible scenarios it introduces, was a little overwhelming, and I didn't mind the feel of what I landed on.

The same is true with the horizontal scrolling in text-based inputs.

I'm definitely open to suggestions or PRs, but I have a few other things on so I probably won't be able to look at this myself for a little while.

jessarcher:
Closing this to continue the discussion at #43

PR #43, opened 2023-08-07
Add view state for scroll prompts

crazywhalecc:
I tried to complete the issue #42 I created earlier, and this PR completed this feature: you can naturally scroll according to the options displayed on the current page, with minimal cost (perhaps).

Untitled

As you know, I added a class to store view state, and changed the scroll behavior.


jessarcher:
I'll need to do a more thorough test and code review, but my first impressions of the UX after checking this out locally are *very* positive! Well done! :raised_hands:

taylorotwell:
@jessarcher mark as ready please when you're ready for me to take a look 👍

jessarcher:
Hey @crazywhalecc!

Sorry for the delay. Your scroll behaviour is perfect, but I've refactored the implementation to better fit my mental model. A side benefit is that we don't need to include those hard-coded 5 and 7 numbers in the Prompt classes, which are theme-specific concerns. It also removes any state manipulation from the renders so they can focus purely on rendering.

Thanks for your PR and for getting this feature happening!

jessarcher (approved):

PR #44, opened 2023-08-08
Prompt `hint`

devajmeireles:
Considering the stylization of the prompt that reminds us of the idea we have in HTML where the typed value is inserted in an input, this PR aims to add `hits` to the inputs: small blocks of text that allow us to give hints, similar to what we do in HTML:

CleanShot 2023-08-08 at 8 28 16

Notes: 
- The hints will be visible as truncated, following the prompt rules.
- The hints will be visible as long as an error message, alert or validation error does not occur.
- The hints is available for all prompt types.

### DIFF
--- a/playground/confirm.php
+++ b/playground/confirm.php
@@ -9,7 +9,8 @@ $confirmed = confirm(
     validate: fn ($value) => match ($value) {
         false => 'You must install dependencies.',
         default => null,
-    }
+    },
+    hint: 'Dependencies are required to run the application.',
 );
 
 var_dump($confirmed);
--- a/playground/multiselect.php
+++ b/playground/multiselect.php
@@ -18,6 +18,7 @@ $permissions = multiselect(
         empty($values) => 'Please select at least one permission.',
         default => null,
     },
+    hint: 'The permissions will determine what the user can do.',
 );
 
 var_dump($permissions);
--- a/playground/password.php
+++ b/playground/password.php
@@ -12,6 +12,7 @@ $password = password(
         strlen($value) < 8 => 'Password should have at least 8 characters.',
         default => null,
     },
+    hint: 'Your password will be encrypted and stored securely.',
 );
 
 var_dump($password);
--- a/playground/search.php
+++ b/playground/search.php
@@ -29,7 +29,8 @@ $model = search(
         if ($value === '0') {
             return 'User 0 is not allowed to receive emails.';
         }
-    }
+    },
+    hint: 'An email will be sent to the user.',
 );
 
 var_dump($model);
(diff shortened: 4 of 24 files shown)


jessarcher:
Thank you! I've been wanting to add this, and I prefer your naming of `hint` :grin: 

I'm going to mark this as draft until I've done a review for Taylor.

jessarcher:
I've made a small tweak to prevent reserving an extra line for errors when a hint is provided because the errors will use that same line.

I've also made the hint appear while searching otherwise it flashes away briefly when the search callback isn't instantaneous.

The only other thing I'm considering is whether to hide the hint when the `search`/`suggest` dropdown is visible rather than the hint jumping down the screen. This would feel a bit more like the web, where a `<select>` dropdown appears in front of the other content on the page, but with the downside that the hint isn't always visible.

devajmeireles:
> I've made a small tweak to prevent reserving an extra line for errors when a hint is provided because the errors will use that same line.
> 
> I've also made the hint appear while searching otherwise it flashes away briefly when the search callback isn't instantaneous.
> 
> The only other thing I'm considering is whether to hide the hint when the `search`/`suggest` dropdown is visible rather than the hint jumping down the screen. This would feel a bit more like the web, where a `<select>` dropdown appears in front of the other content on the page, but with the downside that the hint isn't always visible.

For me, it looks fantastic! I believe so, it will be better to hide the tooltip when `search`/`select` is opened. Do you want me to do this? Let me know if I can help.

devajmeireles:
> > I've made a small tweak to prevent reserving an extra line for errors when a hint is provided because the errors will use that same line.
> > I've also made the hint appear while searching otherwise it flashes away briefly when the search callback isn't instantaneous.
> > The only other thing I'm considering is whether to hide the hint when the `search`/`suggest` dropdown is visible rather than the hint jumping down the screen. This would feel a bit more like the web, where a `<select>` dropdown appears in front of the other content on the page, but with the downside that the hint isn't always visible.
> 
> For me, it looks fantastic! I believe so, it will be better to hide the tooltip when `search`/`select` is opened. Do you want me to do this? Let me know if I can help.

On second thought, I also have no problem keeping the hint for `search`/`select`. Seems to work fine in my tests.

devajmeireles:
@jessarcher , Can I mark the PR as ready for review?  I don't know when to do this. Let me know if I can help with anything else.

jessarcher:
> On second thought, I also have no problem keeping the hint for `search`/`select`. Seems to work fine in my tests.

Agreed - I just tried hiding it when the search/suggest menu is visible and I think I prefer it how it is.

jessarcher (approved):

Issue #45, opened 2023-08-09
Usage of colors makes the right border position incorrect


sc85:
### Laravel Prompts Version

0.1.14

### Laravel Version

10.18.0

### PHP Version

8.1.14

### Operating System & Version

macOS 13.5

### Terminal Application

Terminal

### Description

When using a color within options the right border is not aligned but shifted to the left.
So far, I noticed the issue with select, multiselect and search.
This issue seems somehow similar to #29.

As visible in the screenshot the number of chars the border is shifted seems to be related to the number of chars used to identify the color.

![image](https://github.com/laravel/prompts/assets/4824890/0ab306cb-18c7-4bcc-b8d5-fe1723f62be2)

P.S.: I love the new prompts 😍 

### Steps To Reproduce

```php
<?php

namespace App\Console\Commands;

use Illuminate\Console\Command;

use function Laravel\Prompts\select;

class Dummy extends Command
{
    /**
     * The name and signature of the console command.
     *
     * @var string
     */
    protected $signature = 'dummy';

    /**
     * The console command description.
     *
     * @var string
     */
    protected $description = 'Dummy command';

    /**
     * Execute the console command.
     *
     * @return int
     */
    public function handle(): int
    {
        select(
            'See the issue?',
            [
                'yes' => '<bg=green>yes</>',
                'no' => '<bg=red>no</>',
                'maybe' => '<bg=yellow>maybe</>',
                'dontKnow' => 'don\'t know',
            ],
        );
        return Command::SUCCESS;
    }
}
```

jessarcher:
Thanks, @devajmeireles!

PR #47, opened 2023-08-09
Draw Boxes With All Supported Colors

devajmeireles:
Fixes: #45 

Output:

CleanShot 2023-08-09 at 2 04 08

### DIFF
--- a/src/Themes/Default/Concerns/DrawsBoxes.php
+++ b/src/Themes/Default/Concerns/DrawsBoxes.php
@@ -82,6 +82,6 @@ trait DrawsBoxes
     {
         $text = preg_replace("/\e[^m]*m/", '', $text);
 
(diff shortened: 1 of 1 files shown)


jessarcher:
Thanks! It would be nice if this supported the rest of the syntax, like foreground colors and other text formatting like bold, italic, etc.

I wonder whether Symfony has any methods we can call to strip the text (e.g. for the non-ANSI output version) or whether we can look at how they're processing the text.

jessarcher:
I've just updated the regex to handle more Symfony formatting tags (like bold, underscores, etc.) as well as handling tags with multiple options (e.g. `<fg=blue;bg=red>` and `<options=bold,underscore></>`).

jessarcher (approved):

PR #49, opened 2023-08-15
Remove redundant handling of Symfony-style tags

jessarcher:
This code was from before we used Symfony's output buffer and was just a crude way to maintain the gray "Provider" and "Tag" labels in the `vendor:publish` command.

The Symfony output buffer was added so that we could track trailing newlines in interleaved output, but it had the added benefit of handling the full range of Symfony formatting tags, making this code redundant.

The changes in #47 allow us to ignore these tags for the purposes of calculating padding, where previously they needed to be converted to escape characters before rendering.


jessarcher:
Looks like the code formatting rules changed and I drew the short straw on having them be part of my PR :sweat_smile:

PR #50, opened 2023-08-15
Add a default hint for the `multiselect` prompt

### LINKED ISSUE #41: Unable to select any options in a multi-select prompt on Windows WSL2
### Laravel Prompts Version

0.1.3

### Laravel Version

10.17.1

### PHP Version

8.2.8

Windows 10 Pro - 22H2

### Terminal Application

WSL2 (via VS Code / Windows Terminal)

### Description

Hi, I'm unable to select any options with multi-select prompts.

This is my first experience using the new prompts package by running `breeze:install` on a new verison of Laravel. I want to enable options here (e.g. "Dark mode") but seem unable to do so. I can only submit it with none selected.

image

Windows WSL2, same issue both via VS Code terminal and also directly in Windows Terminal for Ubuntu.

laravel prompts 0.1.3

I have tried several different control and alphanumeric keys on my keyboard, to no effect. I can only navigate up and down between options, but not select them.

In [another issue](https://github.com/laravel/prompts/issues/31), @jessarcher mentioned this:
> Could you try running the following minimal debugging script on the affected command line

This is my result with that script... entering: UP, DOWN, LEFT, RIGHT, h, j, ENTER, Ctrl+C:

```
"\u001b[A"
"\u001b[B"
"\u001b[D"
"\u001b[C"
"h"
"j"
"\n"
"\u0003"
Restoring TTY
```

How do I select an option from this list?

### Steps To Reproduce

Use any artisan command that uses an optional multi-select list of options, e.g. `php artisan breeze:install`

### LINKED ISSUE #48: Add help text or expected input to multi-select prompts
When presented with a multi-select prompt, it's not clear what input is expected in order to select options:

Screenshot 2023-08-11 at 11 00 14 PM

I looked in the [docs](https://laravel.com/docs/10.x/prompts#multiselect) but didn't see any info there. Ultimately I ended up on the issue tracker where another user had the same issue: https://github.com/laravel/prompts/issues/41 I tried various key combinations (y/n, 1/0, return, tab/ctrl/shift/alt/cmd combinations) but spacebar wasn't one of them 🫣. A line that says "Use spacebar to select options" or a mention in the docs (or both) would definitely help here.

@jessarcher You mentioned the possibility of adding this feature https://github.com/laravel/prompts/issues/41#issuecomment-1666808775 so I figured I'd open up an issue to +1 the idea.

jessarcher:
This PR uses the `hint` feature introduced in #44 to add a default hint to the `multiselect` prompt, instructing users to use the space bar to select options. I went with the term "space bar" as it is the dominant wording used on [Wikipedia](https://en.wikipedia.org/wiki/Space_bar).

image

A few people have raised this issue specifically referring to the multi-select prompt used in the Laravel installer, which will be many people's first exposure to Prompts and Laravel. We could just add this hint to the multi-select prompts in the installer, but ultimately it seemed more user-friendly to make it the default. If someone does not want the hint, they would need to pass an empty string to the `hint` argument.

Fixes #41
Fixes #48

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -38,7 +38,7 @@ function select(string $label, array|Collection $options, int|string $default = 
  * @param  array<int|string>|Collection<int, int|string>  $default
  * @return array<int|string>
  */
-function multiselect(string $label, array|Collection $options, array|Collection $default = [], int $scroll = 5, bool|string $required = false, Closure $validate = null, string $hint = ''): array
+function multiselect(string $label, array|Collection $options, array|Collection $default = [], int $scroll = 5, bool|string $required = false, Closure $validate = null, string $hint = 'Use the space bar to select options.'): array
 {
     return (new MultiSelectPrompt($label, $options, $default, $scroll, $required, $validate, $hint))->prompt();
 }


shane-smith:
This is great, thank you @jessarcher!

Issue #51, opened 2023-08-16
Feature Request: Progress Bar


adrianb93:
Amazing work, Jess. `laravel/prompts` is mint 👌

I would love to suggest a progress bar component.

It's not really an input, but it is something you could cancel, or it could be combined with a text input - for example entering a URL to download from.

For example:

- Label
- Cancelled state
- Customise the display value for the numerator/denominator
- Pending mode _(like a loading spinner variant if progress is unknowable)_
- Parallel progress bars? - or maybe parallel output could be a seperate feature of its own
- Maybe `Http` download support, i.e. sugar for `Http::sink('/path/to/file')->withOptions(['progress' => ...])->get($url)`

devajmeireles:
> 

Particularly I think the idea of progress bar is very old. Why not use the [spin](https://github.com/laravel/prompts/blob/main/playground/spin.php) instead? It is enough that when a process finishes, you call another spin, sequentially:

![CleanShot 2023-08-16 at 11 38 26](https://github.com/laravel/prompts/assets/60591772/a6db8a65-dc80-455a-815d-5c57a1ca4e21)

Perhaps something like that would be awesome. I will try to suggest:

```php
spin(fn () => sleep(3), 'Installing dependencies...')
    ->then(fn () => sleep(3), 'Preparing environment...')
    ->then(fn () => sleep(3), 'Finishing up...');
```

adrianb93:
> Why not use the [spin](https://github.com/laravel/prompts/blob/main/playground/spin.php) instead?

I had no idea that spin was in here. It's not documented.

> Particularly I think the idea of progress bar is very old.

Subjective. For most cases I'd agree. For my needs: 

- I have a large file download that I'd prefer to see progress, time elapsed, and time remaining for. 
- I've also worked on import commands that work through many files (an unknown amount) where you'd want to know the progress.

I also think a progress component would look cool for job batches as a use-case for the framework.

PR #54, opened 2023-08-16
Spin with `then` Method

devajmeireles:
According to [this issue](https://github.com/laravel/prompts/issues/51), I had the idea of suggesting an additional resource for `spin`, evaluating the result of the callback return. **When the callback is evaluated to false (none, null, false), we can chain the `then` method in sequence to create a fluent interface**, as demonstrated below:

CleanShot 2023-08-16 at 12 24 45

### DIFF
--- a/playground/spin.php
+++ b/playground/spin.php
@@ -17,4 +17,29 @@ echo PHP_EOL;
 
 var_dump($result);
 
-echo str_repeat(PHP_EOL, 6);
+spin(
+    function () {
+        sleep(2);
+    },
+    'Starting process...',
+)->then(
+    function () {
+        sleep(2);
+    },
+    'Installing dependencies...',
+)->then(
+    function () {
+        sleep(2);
+    },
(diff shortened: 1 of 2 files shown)


jessarcher:
Hey @devajmeireles, thanks for the PR!

It's a neat idea, but I think it messes with the return value too much. With this change, it's no longer possible to return `false` from a single spinner (you'll get the `Spinner` instance instead). Also, once you call `then`, the return values are completely ignored, so you can't return anything nor prevent the next item in the chain from occurring.

Also, the spinner is still quite experimental (and undocumented). I want to learn more about how people want to use it before making too many changes that limit what else we can do. I also have some ideas of my own such as the ability to stream output from a console command to display a scrolling log while the spinner remains at the bottom. We might want to add an API that lets you change the message of an existing spinner while it's running. We also might want to return false to render an error state.

devajmeireles:
> 

Make sense. Thanks to the feedback.

PR #57, opened 2023-08-25
Add default fallbacks

jessarcher:
This PR adds default fallback behavior directly to the Prompts library so that it doesn't _need_ to be implemented by consuming projects and packages.

This will allow us to remove [this code](https://github.com/laravel/installer/blob/master/src/Concerns/ConfiguresPrompts.php) from the Laravel installer.

We could also remove the same code from `laravel/framework`, however, the framework fallback currently uses the updated styling from Nuno.

### DIFF
--- a/src/ConfirmPrompt.php
+++ b/src/ConfirmPrompt.php
@@ -3,6 +3,8 @@
 namespace Laravel\Prompts;
 
 use Closure;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Style\SymfonyStyle;
 
 class ConfirmPrompt extends Prompt
 {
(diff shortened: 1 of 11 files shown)


src/ConfirmPrompt.php
```diff
@@ -49,4 +51,17 @@ public function label(): string
...
         return $this->confirmed ? $this->yes : $this->no;
     }
+
+    /**
+     * Configure the default fallback behavior.
+     */
+    protected function configureDefaultFallback(): void
+    {
+        self::fallbackUsing(fn (self $prompt) => $this->retryUntilValid(
+            fn () => (new SymfonyStyle(new ArrayInput([]), static::output()))->confirm($prompt->label, $prompt->default),
```
jessarcher, line 61: We'd need to call the `setInteractive` method on the `Input` object to respect the `--no-interaction` option.

Implementing a first-party "non-interactive" feature in Prompts is probably worthwhile so we can return an appropriate value where possible and ensure that any `required` and `validate` arguments would pass regardless of the fallback configuration.

Symfony's choice component, for example, does not allow you to select no option but will return `null` in non-interactive mode.

utsavsomaiya:
I don't know it is right or wrong for all?

image

image

But it's need the all has installed the `tput`

jessarcher:
Will revisit this later

utsavsomaiya:
okay!

PR #59, opened 2023-09-08
fix(select): default value not being centered when not visible

toyi:
When you have a `select` prompt and set a default value outside the first x values (where x is defined by the `scroll` parameter), you don't see it. 

It's even worst when you scroll down the list, since you'll never see what is currently selected, it'll remain outside of the displayed values until the end, when it loops back to the top.

This PR fixes this by centering the "view" (the displayed values) on the default item.

I had to modify the `select.php` playground in order to reflect this change.

### Before: 

https://github.com/laravel/prompts/assets/15825155/7f7a6954-27af-43e0-8ff0-f4b21f675643

### After: 

https://github.com/laravel/prompts/assets/15825155/f1036fb2-42bb-4be6-b543-0f245d0441d0

...and when the default value is near the bottom of the list (in this example with a scroll 10):

https://github.com/laravel/prompts/assets/15825155/2405c665-d6c8-4c3f-a485-deba20272995

### DIFF
--- a/src/Concerns/Fallback.php
+++ b/src/Concerns/Fallback.php
@@ -27,6 +27,14 @@ trait Fallback
     }
 
     /**
+     * Disable the fallback implementation.
+     */
+    public static function disableFallback(): void
+    {
+        static::$shouldFallback = false;
+    }
+
+    /**
      * Whether the prompt should fallback to a custom implementation.
      */
     public static function shouldFallback(): bool
--- a/src/Concerns/Fallback.php
+++ b/src/Concerns/Fallback.php
@@ -27,14 +27,6 @@ trait Fallback
     }
 
     /**
-     * Disable the fallback implementation.
-     */
-    public static function disableFallback(): void
-    {
-        static::$shouldFallback = false;
-    }
-
-    /**
      * Whether the prompt should fallback to a custom implementation.
      */
     public static function shouldFallback(): bool
--- a/src/SelectPrompt.php
+++ b/src/SelectPrompt.php
@@ -45,6 +45,16 @@ class SelectPrompt extends Prompt
             } else {
                 $this->highlighted = array_search($this->default, array_keys($this->options)) ?: 0;
             }
+
+            // If the default is not visible, scroll and center it.
+            // If it's too near the end of the list, we scroll to the end.
+            if($this->highlighted >= $this->scroll){
+                $optionsLeft = count($this->options) - $this->highlighted;
+                $halfScroll = floor($this->scroll / 2);
+                $endOffset = max(0, $halfScroll - $optionsLeft);
+
+                $this->firstVisible = $this->highlighted - $halfScroll - $endOffset;
+            }
         }
 
         $this->on('key', fn ($key) => match ($key) {
--- a/src/SelectPrompt.php
+++ b/src/SelectPrompt.php
@@ -50,8 +50,17 @@ class SelectPrompt extends Prompt
             // If it's near the end of the list, we just scroll to the end.
             if($this->highlighted >= $this->scroll){
                 $optionsLeft = count($this->options) - $this->highlighted - 1;
-                $halfScroll = (int) floor(($this->scroll - 0.1) / 2);
+                $halfScroll = (int) floor($this->scroll / 2);
                 $endOffset = max(0, $halfScroll - $optionsLeft);
+
+                // If the scroll is even, we need to subtract one more
+                // in order to take the highlighted option into account.
+                // Since when the scroll is odd the halfScroll is floored,
+                // we don't need to do anything.
+                if($this->scroll % 2 === 0){
+                    $endOffset--;
+                }
+
                 $this->firstVisible = $this->highlighted - $halfScroll - $endOffset;
             }
         }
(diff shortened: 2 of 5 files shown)


src/SelectPrompt.php
```diff
@@ -45,6 +45,15 @@ public function __construct(
             } else {
                 $this->highlighted = array_search($this->default, array_keys($this->options)) ?: 0;
             }
+
+            // If the default is not visible, scroll and center it.
+            // If it's near the end of the list, we just scroll to the end.
+            if($this->highlighted >= $this->scroll){
+                $optionsLeft = count($this->options) - $this->highlighted - 1;
+                $halfScroll = (int) floor(($this->scroll - 0.1) / 2);
```
jessarcher, line 53: I'm curious what the 0.1 solves?

I'm guessing it's to just to make the selected item appear in the top half instead of the bottom half when `scroll` is an even number?

jessarcher:
Good catch! Thanks for fixing it!

Unfortunately, the issue still exists when the terminal height can't accommodate the configured `scroll` value.

It's not straightforward to solve, though. Currently, the renderer reduces the prompt's `scroll` property if needed, considering any additional lines it needs to render above and below the scrolling section. This happens after the prompt is constructed, which is too late.

I'm working on a solution so that renderers can expose the number of lines they need to reserve, and then the prompt itself can adjust the `scroll` property.

src/SelectPrompt.php
toyi: While explaining why I did this, I encountered another add / even bug, so i rewrote this part (removed the confusing `- 0.1`) and made more automated tests on that part.

This is tricky because two things interacts with each other.

First, so we are on the same page, what's the problem? The problem is, when you are near the end of the list, you can't just center the default value because there is not enough items after it. 

For a scroll of 9, you end up with something like:

```
 │   ○ Portugal                                               │ │
 │   ○ South Africa                                           │ │
 │   ○ South Korea                                            │ │
 │   ○ Spain                                                  │ │
 │ › ● Switzerland                                            │ │
 │   ○ United Kingdom                                         │ │
 │   ○ United States                                          │ │
```

So we need an offset, here called `$endOffset`, to compensate the missing items and therefore to keep a list of 9 items:

```
 │   ○ Nigeria                                                │ │
 │   ○ New Zealand                                            │ │
 │   ○ Portugal                                               │ │
 │   ○ South Africa                                           │ │
 │   ○ South Korea                                            │ │
 │   ○ Spain                                                  │ │
 │ › ● Switzerland                                            │ │
 │   ○ United Kingdom                                         │ │
 │   ○ United States                                          │ │
```

The half scroll can't be a `float` so I floor it, 9 becomes 4.5, then 4. This means 4 items before and 4 items after the highlighted value, so 9 in total.

For an even scroll however, the floor does nothing, 10 becomes 5... so 5 items before, 5 items after... and the highlighted items in the center. 11 items in total, in instead of 10.

That's why we need to compensate for this highlighted item and decrement the `$endOffset`. This way, the highlighted item is taken into account and everything works.

You maybe noted the `$endOffset` value can end up being `-1` when we are not near the bottom of the list, which just put the highlighted value in the upper part (which is nicer than the bottom part, but that's just my opinion here).

These problems can quickly make you mad, I totally understand if this is still not 100% clear, especially when you have not worked on it. If you need more explanations, feel free to ask :)

src/Concerns/Fallback.php
toyi, line 1: I added this method because if you don't disable the fallback implementation, all tests that follow the one that used the fallback implementation also use it.

In my case, since the `can fall back` test is using a fallback implementation, my tests failed. Calling `Prompt::disableFallback();` at the end of this test solved the problem.

toyi:
> Good catch! Thanks for fixing it!
> 
> Unfortunately, the issue still exists when the terminal height can't accommodate the configured `scroll` value.
> 
> It's not straightforward to solve, though. Currently, the renderer reduces the prompt's `scroll` property if needed, considering any additional lines it needs to render above and below the scrolling section. This happens after the prompt is constructed, which is too late.
> 
> I'm working on a solution so that renderers can expose the number of lines they need to reserve, and then the prompt itself can adjust the `scroll` property.

Good catch too, I didn't even think about it!

src/Concerns/Fallback.php
jessarcher: I ran into a similar issue in #57. I've pushed a tweak to this PR to let PHPUnit deal with it for us and reset the state.

jessarcher: Oh wait - I don't think you've enabled maintainers to push changes to your branch.

toyi: Hum, strange, it should be allowed. Where should I look?

I didn't do anything unusual, just forked the repo and made the PR.

image

jessarcher: All good - I've pushed my change to `main`. If you rebase/merge you shouldn't need the `disableFallback` method any longer.

jessarcher:
> I'm working on a solution so that renderers can expose the number of lines they need to reserve, and then the prompt itself can adjust the `scroll` property.

I've submitted #60, which makes sure the `$scroll` property is correctly set before your new scroll logic.

toyi:
> I've submitted https://github.com/laravel/prompts/pull/60, which makes sure the $scroll property is correctly set before your new scroll logic.

Thank you, I've updated my main branch and deleted the now useless `disableFallback` method.

jessarcher:
Thanks!

jessarcher:
Somewhat related to this PR (and #58), I've just been playing around with showing how many items have been selected in a multi-select, which is especially helpful when they're not all on-screen.

image

PR #63, opened 2023-09-11
Use fallback implementation when `stty` command fails

### LINKED ISSUE #62: Artisan aroow keys don't work
### Laravel Prompts Version

latest

### Laravel Version

10.22.0

### PHP Version

8.2.10

Windows 10

### Terminal Application

PowerShell with SSH command

### Description

The right and left keys do not work to select the option.
"^[[D" appears for the left key and "^[[C" for the right key, it is not possible to select the desired option.

WindowsTerminal_F2tZgpvOyn

### Steps To Reproduce

- on windows open powershell 
- php artisan migrate

jessarcher:
In scenarios where the `fallbackWhen` condition is false, but the `stty` command fails, the prompt will still attempt to render but is unlikely to work correctly.

This has been reported when SSHing into a Linux machine from a Windows terminal, where the environment detection will report a supported environment, but the `stty` command cannot set the appropriate terminal modes. (Note: This may be due to shared hosting permissions rather than Windows - See https://github.com/laravel/prompts/issues/62#issuecomment-1713052440.)

This PR updates Prompts to enable the fallback when the `stty` command fails. I have configured it to still output the `stty` error message in case it can be resolved, but it will only output it once, before the first prompt.

image

I've also improved the error message when no fallback has been registered for users on old Laravel versions or not using Laravel at all. #57 should improve this.

Fixes #62

### DIFF
--- a/src/Terminal.php
+++ b/src/Terminal.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Prompts;
 
+use RuntimeException;
 use Symfony\Component\Console\Terminal as SymfonyTerminal;
 
 class Terminal
@@ -36,9 +37,9 @@ class Terminal
      */
     public function setTty(string $mode): void
     {
-        $this->initialTtyMode ??= (shell_exec('stty -g') ?: null);
+        $this->initialTtyMode ??= $this->exec('stty -g');
 
-        shell_exec("stty $mode");
+        $this->exec("stty $mode");
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Terminal.php
```diff
@@ -76,4 +77,29 @@ public function exit(): void
...
+
+    /**
+     * Execute the given command and return the output.
+     */
+    protected function exec(string $command): string
+    {
+        $process = proc_open($command, [
+            1 => ['pipe', 'w'],
+            2 => ['pipe', 'w'],
+        ], $pipes);
```
utsavsomaiya, line 89: which is run silently?

jessarcher: Sorry, I don't understand the question.

PR #67, opened 2023-09-13
Support emacs style key binding 

storyn26383:
With this PR, we can use `CTRL_P`, `CTRL_F` `CTRL_N`, `CTRL_B` and `CTRL_H` to do `UP`,  `RIGHT`, `DOWN`, `LEFT` and `BACKSPACE`.

### DIFF
--- a/src/MultiSelectPrompt.php
+++ b/src/MultiSelectPrompt.php
@@ -62,8 +62,8 @@ class MultiSelectPrompt extends Prompt
         $this->reduceScrollingToFitTerminal();
 
         $this->on('key', fn ($key) => match ($key) {
-            Key::UP, Key::UP_ARROW, Key::LEFT, Key::LEFT_ARROW, Key::SHIFT_TAB, 'k', 'h' => $this->highlightPrevious(),
-            Key::DOWN, Key::DOWN_ARROW, Key::RIGHT, Key::RIGHT_ARROW, Key::TAB, 'j', 'l' => $this->highlightNext(),
+            Key::UP, Key::UP_ARROW, Key::LEFT, Key::LEFT_ARROW, Key::SHIFT_TAB, Key::CTRL_P, 'k', 'h' => $this->highlightPrevious(),
+            Key::DOWN, Key::DOWN_ARROW, Key::RIGHT, Key::RIGHT_ARROW, Key::TAB, Key::CTRL_N, 'j', 'l' => $this->highlightNext(),
             Key::SPACE => $this->toggleHighlighted(),
             Key::ENTER => $this->submit(),
             default => null,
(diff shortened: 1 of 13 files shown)


src/MultiSelectPrompt.php
```diff
@@ -62,8 +62,8 @@ public function __construct(
         $this->reduceScrollingToFitTerminal();
 
         $this->on('key', fn ($key) => match ($key) {
-            Key::UP, Key::UP_ARROW, Key::LEFT, Key::LEFT_ARROW, Key::SHIFT_TAB, 'k', 'h' => $this->highlightPrevious(),
-            Key::DOWN, Key::DOWN_ARROW, Key::RIGHT, Key::RIGHT_ARROW, Key::TAB, 'j', 'l' => $this->highlightNext(),
+            Key::UP, Key::UP_ARROW, Key::LEFT, Key::LEFT_ARROW, Key::SHIFT_TAB, Key::CTRL_P, 'k', 'h' => $this->highlightPrevious(),
+            Key::DOWN, Key::DOWN_ARROW, Key::RIGHT, Key::RIGHT_ARROW, Key::TAB, Key::CTRL_N, 'j', 'l' => $this->highlightNext(),
```
jessarcher, line 66: Should this have `CTRL_F` and `CTRL_B` for consistency with the existing left/right/h/l behaviour?

src/SelectPrompt.php
```diff
@@ -70,8 +70,8 @@ public function __construct(
         }
 
         $this->on('key', fn ($key) => match ($key) {
-            Key::UP, Key::UP_ARROW, Key::LEFT, Key::LEFT_ARROW, Key::SHIFT_TAB, 'k', 'h' => $this->highlightPrevious(),
-            Key::DOWN, Key::DOWN_ARROW, Key::RIGHT, Key::RIGHT_ARROW, Key::TAB, 'j', 'l' => $this->highlightNext(),
+            Key::UP, Key::UP_ARROW, Key::LEFT, Key::LEFT_ARROW, Key::SHIFT_TAB, Key::CTRL_P, 'k', 'h' => $this->highlightPrevious(),
+            Key::DOWN, Key::DOWN_ARROW, Key::RIGHT, Key::RIGHT_ARROW, Key::TAB, Key::CTRL_N, 'j', 'l' => $this->highlightNext(),
```
jessarcher, line 74: Same here re: `CTRL_F` and `CTRL_B`

jessarcher (commented):
Very cool!

Makes me wonder whether we should add some readline keys like `CTRL_A` and `CTRL_E`, as well as `Home` and `End`. They'd probably need to work differently depending on whether you're currently focused on a list or a typed input. Someone can PR that separately if you don't feel like it belongs here though.

src/MultiSelectPrompt.php
storyn26383: Sure! I lost it, sorry about that.

storyn26383:
Good idea!
We can add home and end feature in next RP 🎉

PR #71, opened 2023-09-15
[10.x] `spin`: Ensure process termination on `$callback()` failure

manelgavalda:
When executing the `spin` function, if an error is thrown within the `$callback()` function, PHP currently displays the error in the terminal but doesn't terminate the process. This pull request introduces a change to ensure that the process is terminated in the event of a failure within the `callback()` function.

### Example:

```php
$result = spin(
    function () {
        sleep(-1);  // This will fail, causing an exception to be caught, and preventing the posix_kill function from being called

        return 'Callback return';
    },
    'Installing dependencies...',
);

### DIFF
--- a/src/Spinner.php
+++ b/src/Spinner.php
@@ -55,6 +55,8 @@ class Spinner extends Prompt
             exit();
         });
 
+        $pid = 0;
+
         try {
             $this->render();
 
--- a/src/Spinner.php
+++ b/src/Spinner.php
@@ -82,6 +82,7 @@ class Spinner extends Prompt
             }
         } catch (\Throwable $e) {
             $this->showCursor();
+            posix_kill($pid, SIGHUP);
             pcntl_async_signals($originalAsync);
             pcntl_signal(SIGINT, SIG_DFL);
 

### CODE REVIEW

File: src/Spinner.php [hunk 1]
```diff
@@ -71,24 +73,32 @@ public function spin(Closure $callback): mixed
...
-                $this->showCursor();
+                $this->resetTerminal($originalAsync, $pid);
 
                 return $result;
             }
         } catch (\Throwable $e) {
-            $this->showCursor();
-            pcntl_async_signals($originalAsync);
-            pcntl_signal(SIGINT, SIG_DFL);
+            $this->resetTerminal($originalAsync, $pid);
```
manelgavalda, line 81: It should do the same as above when in it fails

src/Spinner.php
```diff
@@ -55,6 +55,8 @@ public function spin(Closure $callback): mixed
             exit();
         });
 
+        $pid = 0;
```
manelgavalda, line 58: Inititalize the `$pid` before the `try` in case `$this->render()` fails.

jessarcher: `0` has a special meaning when passed to `posix_kill`:

image

I'm not entirely sure of the implications, so I think it would be preferable to remove this line and handle it not being set.

src/Spinner.php
```diff
@@ -71,24 +73,29 @@ public function spin(Closure $callback): mixed
...
-                $this->showCursor();
+                $this->resetTerminal($originalAsync, $pid);
 
                 return $result;
             }
         } catch (\Throwable $e) {
-            $this->showCursor();
-            pcntl_async_signals($originalAsync);
-            pcntl_signal(SIGINT, SIG_DFL);
+            $this->resetTerminal($originalAsync, $pid);
```
jessarcher, line 81:
```suggestion
            $this->resetTerminal($originalAsync, $pid ?? null);
```

src/Spinner.php
```diff
@@ -71,24 +73,29 @@ public function spin(Closure $callback): mixed
...
 
             throw $e;
         }
     }
 
+    /**
+     * Reset the terminal.
+     */
+    protected function resetTerminal(bool $originalAsync, int $pid): void {
+        posix_kill($pid, SIGHUP);
```
jessarcher, line 91:
```suggestion
    protected function resetTerminal(bool $originalAsync, ?int $pid): void {
        if ($pid) {
            posix_kill($pid, SIGHUP);
        }
```

src/Spinner.php
manelgavalda: That's right, thank you. Just updated the PR with the changes :+1:

jessarcher (approved):

PR #72, opened 2023-09-17
Remove extra line `spin` leaves behind

joetannenbaum:
I found that when using `spin`, it leaves an extra line behind, creating a gap in output. Not sure if it's intentional, but in this PR I moved the cursor up one line before returning the result of the callback.

Given the following code:

```php
text('First question');
text('Second question');

$result = spin(
    function () {
        sleep(4);

        return 'Callback return';
    },
    'Installing dependencies...',
);

info($result);
info('Also this!');
```

You would get the resulting output:

Before:

CleanShot 2023-09-17 at 14 15 30@2x

After:

CleanShot 2023-09-17 at 14 15 09@2x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Release Notes
 
-## [Unreleased](https://github.com/laravel/prompts/compare/v0.1.7...main)
+## [Unreleased](https://github.com/laravel/prompts/compare/v0.1.8...main)
+
+## [v0.1.8](https://github.com/laravel/prompts/compare/v0.1.7...v0.1.8) - 2023-09-19
+
+- Add a `table` renderer by [@joetannenbaum](https://github.com/joetannenbaum) in https://github.com/laravel/prompts/pull/68
+- Support emacs style key binding  by [@storyn26383](https://github.com/storyn26383) in https://github.com/laravel/prompts/pull/67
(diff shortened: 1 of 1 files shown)


jessarcher:
Hey @joetannenbaum, thanks for the PR!

This one is a little complicated. I can confirm the issue you've demonstrated, but there are some different issues with this solution. For example, if you put a spinner as the first output, or put two spinners in a row, then it erases too much.

It all has to do with how each prompt determines how many newlines to output above the prompt, which involves looking at the trailing newlines previously output on the buffer (and also handling the case where nothing was previously output).

Ultimately it would be great if the spinner could restore the console output to how it was and also restore the previous trailing newline count so the next prompt calculates things as though the spinner was never there, rather than considering whatever trailing newlines the spinner output before being erased.

jessarcher:
Feel free to let me know if you have trouble, and I'll see if I can help.

I have a branch at https://github.com/laravel/prompts/compare/debug-markers?expand=1 that allows you to visually see newline characters and render cycle counts. It tries not to get in the way of the output, but I think results in an extra line after the command finishes because of the trailing render count on the final line. Debugging with Ray is sometimes easier to see what's happening without impacting the console output.

joetannenbaum:
Awesome! Thanks for the tip on the debug branch, super useful. I'll keep you posted 👍

Issue #17, opened 2023-07-29
Porting interactive functionality to Symfony


wouterj:
I asked about this on [Twitter last week](https://twitter.com/wouterjnl/status/1681991895998689282), but I realize that is not the best place to discuss this. I'm really digging the interactive prompts (arrow-based choice lists & completion dropdowns) of this library! This is something I was trying to introduce to Symfony Console for many years (but lacking the stty skills needed for this).

Would it be an idea to see if this part of Prompts can be ported back to Symfony Console, a library you already depend on in the Laravel ecosystem and this library? That way, a larger part of the PHP ecosystem can have this much better terminal UX (including tooling used in the Laravel ecosystem like Composer and PHPstan).

The way I see it, this functionality is a perfect fit for the [Question classes in Symfony Console](https://github.com/symfony/console/tree/6.3/Question). Laravel Prompts can then use the question classes, and add the custom "Laravel style" (plus any other extra integration, like the helper functions) on top of it.

Of course, given the code in this library is MIT, I can always try this proposal without your consent. But I'm looking forward to hear your ideas about this (e.g. did you try this and reached a major blocker already, is this just "not worth it" for your time, are you happy to try this out, or can I stalk a bit when experimenting with this myself :grin: )

jessarcher:
Hey @wouterj! Sorry I missed your Twitter post - thank you for sharing it!

I would love to see something like this in Symfony. The challenge is supporting native Windows (i.e. `php.exe` without WSL) and potentially other environments. Currently, we're configuring the Prompts in Laravel to [fall back to Symfony in Windows](https://github.com/laravel/framework/pull/46772/files#diff-9c0b99959f28e3b7a54a93a4ff67818de374a378726f8ecd16fa3c3bded6a9c7R27) because I couldn't find a way to make the arrow keys work, amongst other things.

I've briefly explained the challenges I faced with Windows at https://github.com/laravel/docs/pull/8924#issuecomment-1654762739.

In Symfony, it's already possible to use the arrow keys as an alternative mode for selection. However, as the entries can still be typed manually, it's still useable with the arrow or tab keys (albeit more frustratingly). After trying to implement Laravel Prompts in Windows, I gained a new understanding and appreciation for why Symfony's prompts are the way they are. It appears they have made an admirable decision to sacrifice some usability in exchange for supporting more environments.

Unless we can figure out how to solve the problems mentioned in my linked comment, I think the only alternative would be exploring alternative UI approaches that allow for graceful degradation.

[...]

PR #73, opened 2023-09-18
[0.2.x] Add non-interactive mode

### LINKED ISSUE #64: Prompts does not check for a TTY
### Laravel Prompts Version

0.1.6

### Laravel Version

10.22.0

### PHP Version

8.2.10

Linux 6.4.6

### Terminal Application

N/A, this is about TTYless enviroments and thus no terminal

### Description

Prompts does zero checking if standard input is a tty. Without an interactive tty, which can be caused running commands from cron or in Docker without allocating one, this causes `stty` [here](https://github.com/laravel/prompts/blob/59df38332c5392b10a805f97646b5c5491862486/src/Terminal.php#L42) to fail with `stty: standard input: Not a tty`. Since there is zero checks to see if this was successful or checking that it's on a tty separately, the code always assumes it is on a tty. When [this while loop](https://github.com/laravel/prompts/blob/59df38332c5392b10a805f97646b5c5491862486/src/Prompt.php#L101) is entered, there are no time out checks to break out of the loop, nor is there a there something to slow it down since `fread(STDIN, 1024)` [here](https://github.com/laravel/prompts/blob/59df38332c5392b10a805f97646b5c5491862486/src/Terminal.php#L28) instantly returns zero length string with out a tty (assuming nothing has been piped in there, or has otherwise been drained).

This has the following consequences, assuming that scheduled commands in cron trigger this, 100% CPU thread utilization per instance started and slow memory exhaustion as concurrently running commands pile up.

There are a few ways I can see fixing this, from detecting the lack of a TTY & throwing an exception to end everything, to accepting default options. My assumption is detect and throw is the better one, but I'm happy to be proven wrong.

### Steps To Reproduce

1. Add a command that uses ` Illuminate\Console\ConfirmableTrait` to the scheduler in Laravel, while forgetting to set `--force` for production
   * Alternatively, upgrade a package you already have a scheduled command for that migrated to using `laravel/prompts`, and not realize the consequences that change causes.
2. Wait for schedule to execute the command without a TTY.
   * Alternatively, you can run a command in your shell this way to execute it without a TTY attached: `(setsid php artisan activitylog:clean) </dev/null`
   * Yes, the parentheses are needed, use of the subshell avoids `setsid` forking and going into the background, and the `/dev/null` redirect is to remove keyboard input, like it would be while cron runs.
3. Profit!

jessarcher:
This PR introduces a non-interactive mode for Prompts.

When Prompts are non-interactive, they will attempt to return the configured default value instead of prompting. Most prompts have a safe default when none is provided. An exception will be thrown if the default value does not pass any validation checks.

This is useful when prompts occur in a non-interactive environment, such as a scheduled task, queued job, CI, or deploy script.

The behaviour is currently mixed, depending on the scenario. #63 prevented prompts from waiting indefinitely for input when the fallback was not triggered, but everything now depends on the fallback to handle non-interactive mode. When using a Symfony fallback, there are some inconsistencies. For example, the `choice` component doesn't allow you to select nothing in interactive mode but will return `null` in non-interactive mode. With this PR, the fallbacks no longer need to be used for non-interactive mode - just for Windows and when testing (for now).

This is a breaking change because Laravel depends on the fallback when running tests and the non-interactive mode will take over in a CI testing environment. If we release this change in `v0.2`, Laravel can update its version constraint and ensure interactive mode is enabled when testing. The companion PR is at https://github.com/laravel/framework/pull/48468

In the future, I would like to prevent the need for fallbacks when testing so that fallback mode is only required for Windows environments (and maybe that can even be solved with #17).

This also paves the way for #57, which needs to know when Laravel commands are explicitly told to be non-interactive (via the `-n` argument), regardless of the presence of a tty.

Fixes #64

### DIFF
--- a/src/Prompt.php
+++ b/src/Prompt.php
@@ -16,6 +16,7 @@ abstract class Prompt
     use Concerns\Events;
     use Concerns\FakesInputOutput;
     use Concerns\Fallback;
+    use Concerns\Interactivity;
     use Concerns\Themes;
 
     /**
@@ -73,6 +74,12 @@ abstract class Prompt
      */
     public function prompt(): mixed
     {
+        static::$interactive ??= stream_isatty(STDIN);
+
+        if (! static::$interactive) {
+            return $this->default();
+        }
+
         $this->capturePreviousNewLines();
 
         if (static::shouldFallback()) {
--- a/src/Prompt.php
+++ b/src/Prompt.php
@@ -311,7 +311,7 @@ abstract class Prompt
     {
         $this->validated = true;
 
-        if (($this->required ?? false) && ($value === '' || $value === [] || $value === false)) {
+        if (($this->required ?? false) && ($value === '' || $value === [] || $value === false || $value === null)) {
             $this->state = 'error';
             $this->error = is_string($this->required) ? $this->required : 'Required.';
 
--- a/src/Concerns/FakesInputOutput.php
+++ b/src/Concerns/FakesInputOutput.php
@@ -16,6 +16,9 @@ trait FakesInputOutput
      */
     public static function fake(array $keys = []): void
     {
+        // Force interactive mode when testing because we will be mocking the terminal.
+        static::interactive();
+
         $mock = \Mockery::mock(Terminal::class);
 
         $mock->shouldReceive('write')->byDefault();
--- a/src/Concerns/Interactivity.php
+++ b/src/Concerns/Interactivity.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Laravel\Prompts\Concerns;
+
+use Laravel\Prompts\Exceptions\NonInteractiveValidationException;
+
+trait Interactivity
+{
+    /**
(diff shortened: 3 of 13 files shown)


src/Prompt.php
```diff
@@ -73,6 +74,12 @@ abstract public function value(): mixed;
      */
     public function prompt(): mixed
     {
+        static::$interactive ??= stream_isatty(STDIN);
```
jessarcher, line 77: This will break existing Laravel applications when testing console commands in a CI environment. They won't fall back to the Symfony implementation, which is required for the test assertions to work.

It can be solved with a small change in the framework, but that won't help if users have this version of prompts with a slightly older version of the framework.

The solution might be:
* Release this as a breaking change (v0.2.0) to prevent current Laravel installations from installing it.
* Update the minimum required version of Prompts in the framework and add code to set interactivity mode when required.

Alternatively, we could default this to `true` for the time being to maintain existing behaviour and then make the breaking change later.

PR #76, opened 2023-09-20
Register shutdown function to kill process in the event of an early exit

joetannenbaum:
I preface this PR by saying I don't _love_ this solution, but I was struggling to find another way to handle this.

There seems to be an issue where if the callback function exits early, the process rendering the spinner continues running:

```php
$result = spin(
    function () {
        exit('Heading out early.');
    },
    'Installing dependencies...',
);
```

https://github.com/laravel/prompts/assets/2702148/601fd315-1f9c-4d1a-ba61-aa748f98b60a

I'm using `register_shutdown_function` for the `$pid` that's not `0`, which, if this is the solution, might negate the need for the `posix_kill` in the `resetTerminal` method.

Is this something you're interested in handling? If so... is there a better way? I couldn't figure out a way to catch this otherwise.

### DIFF
--- a/src/Spinner.php
+++ b/src/Spinner.php
@@ -71,12 +71,14 @@ class Spinner extends Prompt
 
                 $result = $callback();
 
-                $this->resetTerminal($originalAsync, $pid);
+                posix_kill($pid, SIGHUP);
+
+                $this->resetTerminal($originalAsync);
 
                 return $result;
             }
         } catch (\Throwable $e) {
-            $this->resetTerminal($originalAsync, $pid ?? null);
+            $this->resetTerminal($originalAsync);
 
             throw $e;
         }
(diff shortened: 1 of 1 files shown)


jessarcher:
Hey @joetannenbaum,

Thanks for this! I think we'll also want to restore the cursor like we do when receiving `SIGINT`.

I'm also wondering whether we should erase the rendered lines on `exit` (and potentially even on `SIGINT`).

jessarcher:
> I'm using `register_shutdown_function` for the `$pid` that's not `0`, which, if this is the solution, might negate the need for the `posix_kill` in the `resetTerminal` method.

I think we'll still need this to stop the renderer process when everything is successful.

jessarcher:
I've made sure the cursor is restored on shutdown in the same way we do with other prompts, where it's a no-op if the cursor is already shown. This is important because we can't unregister the shutdown function.

I've also simplified the killing of the child process from #71 now that your handler will take care of that too. I've tested that your shutdown handler is a no-op when the spinner doesn't error out.

jessarcher (approved):

joetannenbaum:
Good call on the cursor restore, didn't think about that. Glad this worked out!

PR #81, opened 2023-09-21
Replace shutdown handlers with destructors

jessarcher:
This PR fixes a memory leak by:

* Replacing calls to `register_shutdown_handler` with class destructors. (thanks @nunomaduro :raised_hands:)
* Clearing the event listeners to remove any circular references in callback functions that would prevent GC.

I've also removed redundant calls to restore the cursor and TTY now that the destructor will handle it immediately after each prompt, regardless of whether it's successful or fails due to exception, exit, or `SIGINT`/`Ctrl+C`.

_Best viewed with "Hide whitespace" enabled._

### DIFF
--- a/src/Spinner.php
+++ b/src/Spinner.php
@@ -23,6 +23,11 @@ class Spinner extends Prompt
     public bool $static = false;
 
     /**
+     * The process ID after forking.
+     */
+    protected int $pid;
+
+    /**
      * Create a new Spinner instance.
      */
     public function __construct(public string $message = '')
@@ -42,8 +47,6 @@ class Spinner extends Prompt
     {
         $this->capturePreviousNewLines();
 
-        register_shutdown_function(fn () => $this->restoreCursor());
-
         if (! function_exists('pcntl_fork')) {
             return $this->renderStatically($callback);
         }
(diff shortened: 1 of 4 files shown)


src/Spinner.php
```diff
@@ -148,4 +145,16 @@ protected function eraseRenderedLines(): void
...
     }
+
+    /**
+     * Clean up after the spinner.
+     */
+    public function __destruct()
+    {
+        parent::__destruct();
+
+        if (! empty($this->pid)) {
```
nunomaduro, line 156: Seems to me, that you want to do `if ($this->pid > 0)`?

src/Spinner.php
```diff
@@ -148,4 +145,16 @@ protected function eraseRenderedLines(): void
         $this->moveCursor(-999, -count($lines) + 1);
         $this->eraseDown();
     }
+
+    /**
+     * Clean up after the spinner.
+     */
+    public function __destruct()
+    {
+        parent::__destruct();
```
nunomaduro, line 154: While in `__constructs` the parent call should comes first, in `__desctruct`, the parent call should be the last line.

nunomaduro (changes requested):

src/Spinner.php
jessarcher: TIL, thanks!

src/Spinner.php
jessarcher: The destructor can be called before the fork occurs, so this property would not be initialised yet.

`isset($this->pid) && $this->pid > 0` would work, but at that point, it's effectively the same as `! empty($this->pid)` given that `$this->pid` is an int.

PR #85, opened 2023-09-23
Stream output while displaying spinner

joetannenbaum:
This PR introduces the optional streaming of output while displaying the spinner. Admittedly, it's still a WIP, but I think it's ready enough to get some more eyeballs on it. This is a fresh PR spawned from the conversation in #77. 

It includes:

- The ability to stream output from a process to the terminal while the spinner is showing
- The ability to change the spinner message as the task is processing

Examples (included a new playground, `playground/streaming-spinner.php`):

```php
spin(
    function (SpinnerMessenger $messenger) {
        $process = Process::fromShellCommandline('composer update');
        $process->start();

        foreach ($process as $type => $data) {
            if ($process::ERR === $type) {
                $messenger->output($data);
            }
        }

        return 'Callback return';
    },
    'Updating Composer...',
);

spin(
    function (SpinnerMessenger $messenger) {
        foreach (range(1, 50) as $i) {
            $messenger->line("<info>✔︎</info> <comment>Step {$i}</comment>");

            usleep(rand(50_000, 250_000));

            if ($i === 20) {
                $messenger->message('Almost there...');
            }

            if ($i === 35) {
                $messenger->message('Still going...');
            }
        }
    },
    'Taking necessary steps...',
);
```

https://github.com/laravel/prompts/assets/2702148/e4b21ac2-a1eb-4d67-92cc-215f34f66fd2

Issues I'm aware of/concerns I have:

- Spacing issues: Two streaming spinners in a row don't have a space between them, a streaming spinner that occurs right away doesn't have a leading space, a streaming spinner that occurs last doesn't have a trailing space
- I don't love that we are using the `Colors` trait and writing the streaming output directly in the `Spinner` class as opposed to a renderer class
- `SpinnerMessenger` isn't the clearest name for the object that is passed as an argument to the task callback, open to suggestions
- I grabbed the `Connection` class directly from spatie/fork, not sure what the general ethics/vibes around that are. It's a really simple class, but just wanted to be above board.
- I haven't included any tests just yet as I'm still finalizing the API

Let me know what you think! Excited about this possibility.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [8.1, 8.2]
+        php: [8.1, 8.2, 8.3]
 
     name: PHP ${{ matrix.php }}
 
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
         "phpstan/phpstan": "^1.10",
         "pestphp/pest": "^2.3",
         "mockery/mockery": "^1.5",
-        "phpstan/phpstan-mockery": "^1.1"
+        "phpstan/phpstan-mockery": "^1.1",
+        "spatie/ray": "^1.39"
     },
     "conflict": {
         "illuminate/console": ">=10.17.0 <10.25.0",
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,10 @@
         "mockery/mockery": "^1.5",
         "phpstan/phpstan-mockery": "^1.1"
     },
+    "conflict": {
+        "illuminate/console": ">=10.17.0 <10.25.0",
+        "laravel/framework": ">=10.17.0 <10.25.0"
+    },
     "suggest": {
         "ext-pcntl": "Required for the spinner to be animated."
     },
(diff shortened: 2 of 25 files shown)


joetannenbaum:
Cleaned this up a bit and got some more consistent output(ish). Right now if you render two streaming spinners back to back, it's consistent. 

But if you render a non-spinner before the spinner, the spacing between the output and the line changes, and it also doesn't erase the line at the end.

Brain is a bit mushy at this point, so I'll come back to this later and see if I can clean it up further.

joetannenbaum:
@jessarcher I think this is ready for review.

I know this PR is a bit of a beast, so no rush here. The whole thing is open to conversation and adjustment, I'm not married to the API or implementation. Also, if you decide it just overcomplicates this component, I would also fully understand.

All that being said, in my testing it displays consistently and looks good on my end.

I tried to write tests for it, but realized that all of the streaming output is actually happening in the child process of the fork, so I couldn't figure out how to access that for the test. That's... probably a smell. 

If you want to take a look and see if you even like the existing implementation I can work on getting tests that work properly.

driesvints:
@joetannenbaum is this still ready for review? Can you mark this as such if so?

joetannenbaum:
I still like this idea, but I'm going to close this PR for now. I'd like to take a fresh spin on it in a future PR (pun intended), try to get it a bit cleaner.

PR #88, opened 2023-09-26
Add Component: Multiline Text Input

joetannenbaum:
This PR introduces a multiline text input called `textarea`. It's parameters and behavior are largely identical to the `text` prompt, with the added parameter of `$rows` to specify how tall the input should be.

https://github.com/laravel/prompts/assets/2702148/ae3a3c49-d576-4471-a152-9ceea9c2b2f1

I really tried to refine this and make it as ready as possible, but I have a couple of "magic numbers" in here that may need more context from @jessarcher to properly implement.

Let me know what you think!


jessarcher:
Wow! This is wild!

I'll check this out soon. A few initial questions:

* Have you considered using `Shift+Enter` either to add a break or submit the message? Not sure it's even possible to read a unique key for `Shift+Enter`.
* What are your thoughts on making it automatically grow in height until it reaches a max height? À la https://twitter.com/yisibl/status/1704879401710846429. It may not be a good idea as it might not be clear that you can enter multiple lines.

joetannenbaum:
As far as I could tell, I wasn't able to detect `Shift+Enter`. I really wanted to get away from the `Ctrl+D` convention as I find that really unnatural, but I couldn't figure out another detectable key combo that wouldn't potentially be typed in the prompt itself. But if you have any ideas I'm all ears.

I initially went with the auto grow scenario (starting with a minimum height of several rows to indicate that it accepted multiline input). Went back and forth on it, I found it to be a little unwieldy when it start getting close to the height of the terminal, so I went with the scrollbar. 

Maybe there's a middle ground there, we let it auto grow, but the max height is something like half of the terminal height (or some other calculation)?

jessarcher:
Hey @joetannenbaum!

I'm sorry for not getting back to you sooner on this one.

I've just merged in the latest changes from `main` and updated the scrolling trait that changed in #79.

I also [fixed](https://github.com/laravel/prompts/pull/88/commits/d513b902f5f5a41bf3ed821274ed323f6a5fac3e) a minor issue when trying to move back to the last line when it's empty.

src/TextareaPrompt.php
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * Get the formatted lines of the current value.
+     *
+     * @return array<int, string>
+     */
+    public function lines(): array
+    {
+        // TODO: Figure out the real number here, this comes from the renderer?
+        $value = wordwrap($this->value(), 59, PHP_EOL, true);
```
jessarcher, line 188: Any ideas on this one? I'm thinking it would be something similar to what we do with the `scroll` property to make sure it will fit on the screen.

joetannenbaum: I took a cue from how you were handling the spinner interval and set a public property on the `Textarea` class from the renderer once the proper width was calculated. I think this solves it if you are good with that, tried it out at a bunch of different sizes and it looked good to me.

joetannenbaum:
@jessarcher I removed the two "magic numbers" I had and replaced them with proper solutions.

Let me know what you think about how I altered the `addCursor` method in `TypedValue`. I made it so that if you pass in a negative value such for the `$maxWidth` argument, it avoided truncating the text altogether.

Felt ok to me as the alternatives would have been to either pass an additional boolean argument or to create another method for getting the value with a cursor sans truncation, which felt like too much duplication. 

Definitely open to suggestions if you're not a fan of the approach.

ClaraLeigh:
How are we progressing on this PR? I was needing a textarea field for a project and would love to start using this

jessarcher:
Hey @ClaraLeigh - things have been pretty hectic with Pulse and Laracon, but I'll be circling back to new Prompts features soon!

jessarcher:
Finally revisiting this!

I've found two obscure issues. I'll try to fix them but documenting them here also:

1. When writing a long line with no spaces and it wraps, the virtual cursor position on the following line is one position to the left. If you continue typing it until it wraps again, the cursor will continue moving to the left for each wrapped line:

   image

2. When writing enough text that it scrolls, when you get to the end of the next line, a weird color inversion occurs:

   image

   I believe this is caused by the virtual cursor being trimmed before it can restore the inversion. When dumping the raw output via Ray, you can see on the second last line that it outputs a `[7m` (invert) but only a `[27` without the trailing `m`, which would undo the inversion.

   image

   If you continue typing onto the next line, it gets more messed up, presumably because the inversion is still "active" when it redraws the whole prompt:

   image

These issues may be related as both seem like an "off by one" error.

I also think that the box should expand to the edge of the cursor before it wraps, similar to how the other prompts work. That's an easy change, but it doesn't seem to fix the above issue.

joetannenbaum:
@jessarcher great catches! Thank you.

I believe I fixed the cursor position for longer words. When the user types longer words, the `typedValue` falls out of sync with the display, since `wordwrap` injects a line break mid-word. I added a method that looks for longer words that wrap to calculate an offset that we take into account when passing the `cursorPosition` into `valueWithCursor`.

I'll add some tests later to triple verify, but on my end it looks good. I'll take a peek at that weird inversion behavior later as well.

jessarcher:
Thanks @joetannenbaum,

There's another issue - PHP's `wordwrap` function isn't multi-byte safe (it will split in the middle of multi-byte characters), and there's no `mb_` equivalent. It also uses length, rather than width, so even if it were multi-byte safe, lines containing full-width characters (such as emojis) would be too wide. I've tried creating an appropriate function using a combination of `mb_strwidth` and `mb_strimwidth`, but there are a lot of edge cases.

jessarcher:
I played around creating a multi-byte safe `wordwrap` function that uses width instead of length and got something that seems to work. I haven't tried putting it all back together with the virtual cursor yet though.

Here's the code as global functions in case it helps:

```php
function mb_wordwrap(string $string, int $width = 75): string
{
    $pieces = array_reduce(
        preg_split('/\b|(\n)/', $string, flags: PREG_SPLIT_DELIM_CAPTURE|PREG_SPLIT_NO_EMPTY),
        fn ($carry, $chunk) => [...$carry, ...mb_str_split_width($chunk, $width)],
        []
    );

    $output = [];
    $line = '';

    foreach ($pieces as $piece) {
        if ($piece === PHP_EOL) {
            $output[] = rtrim($line);
            $line = '';
        } elseif (mb_strwidth(rtrim($line.$piece)) <= $width) {
            $line .= $piece;
        } else {
            $output[] = rtrim($line);
            $line = $piece;
        }
    }

    if ($line !== '') {
        $output[] = rtrim($line);
    }

    return implode(PHP_EOL, $output);
}

function mb_str_split_width(string $string, int $width): array
{
    $chunks = [];

    while (mb_strwidth($chunk = mb_strimwidth($string, $start ??= 0, $width)) !== 0) {
        $chunks[] = $chunk;
        $start += mb_strlen($chunk);
    }

    return $chunks;
}
```

I've mimicked the behaviour of a real `<textarea>` where white space at the end of the line doesn't cause a wrap until non-blank character is typed, but I think the `rtrim` will cause issues navigating with the cursor. The behaviour may need to change to better work with a terminal.

Also, the length can change quite drastically (and could even get shorter due to the trimming), so the only way I can think to place the virtual cursor would be to add a placeholder before wrapping and then replace it afterwards, but the placeholder would need to be the same width to not mess up the calculations. I'm not sure how to handle moving up and down either.

joetannenbaum:
Ok, interesting. I took a different approach, where I just tried to match the behavior of `wordwrap` and then make it as multi-byte safe as possible. I'm sure there are edge cases I'm missing but this was my stab at it:

https://github.com/joetannenbaum/multi-byte-wordwrap/blob/main/src/multi-byte-wordwrap.php

Also, I'm wondering: Do we want to cap the width at a more comfortable reading width instead of expanding the full width of the terminal? Something around 70-75 characters? I think expanding to the entire width makes sense for a single input, but for multi-line it may get a little unwieldy? Just a thought.

https://baymard.com/blog/line-length-readability

joetannenbaum:
Ok, I got this a bit further along. It feels pretty consistent now, and from what I can tell those edge cases have been cleaned up. I went with my `mb_wordwrap` implementation just for now, if you'd like to use yours I'm totally down, we'll just have to re-work the cursor stuff a bit.

Let me know what you think, I think it's feeling pretty good overall.

src/TextareaPrompt.php
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * Get the formatted lines of the current value.
+     *
+     * @return array<int, string>
+     */
+    public function lines(): array
+    {
+        // Subtract 2 for the scrollbar
+        $this->maxLineWidth = $this->width - 2;
```
jessarcher, line 194: Can we subtract 2 from `width` in the renderer and remove the `maxLineWidth` property? The scrollbar feels like a renderer concern. It looks like we're using the current value of `width` in one place in this file, though, and I'm not sure of the consequences of it being reduced by 2.

There is a tiny issue that might be related. If you enter the following:

```
a
bb
ccc
```

With the cursor at the very end, if you press the `Up` arrow, it will move to the beginning of the line instead of going up to the end of the previous line. It only seems to happen when the current line is precisely one character longer than the previous. If it's shorter, equal, or more than one character longer, then it moves up as expected.

jessarcher (commented):
This is looking great! I like your `mb_wordwrap` implementation too.

joetannenbaum:
Sorry for the massive delay in working on this, finally got a moment tonight. 

I think I figured out the weird bug with the up arrow cursor, there was actually a correlating bug with the down arrow as well. Looks like I may have reversed some logic at this point, it looks good to me now.

I also fixed a bug that I encountered where the cursor wasn't in the scrolling viewport anymore when pasting large chunks of text that spanned multiple lines.

Overall in playing with it tonight I think it's looking pretty good!

joetannenbaum:
Yes, this is still active. I'll take a look at what's failing today and fix 👍

driesvints:
@joetannenbaum tests still fail here.

taylorotwell:
Just waiting on final thumbs up from @jessarcher here.

src/helpers.php
```diff
@@ -192,3 +200,89 @@ function progress(string $label, iterable|int $steps, ?Closure $callback = null,
 
     return $progress;
 }
+
+/**
+ * Multi-byte version of wordwrap.
+ *
+ * @param  non-empty-string  $break
+ */
+function mb_wordwrap(
```
jessarcher, line 209: Wondering whether it's better to encompass this in a method on the `TextareaPrompt` class to prevent it showing up in IDE completion.

src/helpers.php
```diff
@@ -13,6 +13,14 @@ function text(string $label, string $placeholder = '', string $default = '', boo
     return (new TextPrompt(...func_get_args()))->prompt();
 }
 
+/**
+ * Prompt the user for multiline text input.
+ */
+function textarea(string $label, int $rows = 5, string $placeholder = '', string $default = '', bool|string $required = false, ?Closure $validate = null, string $hint = ''): string
```
jessarcher, line 19: Wondering if the `$rows` param should be moved to after the `$default` param (the same position as the similar `$scroll` parameter on `select` and `multiselect`)

jessarcher (commented):
Thanks @joetannenbaum! Found one more unexpected behaviour - if you type the following:

```
abc
de
f
```

Then, place the cursor at the end of the first line (after the `c`) and press the down key. The cursor lands on the `f` when I would expect it to land after the `e`.

joetannenbaum:
@jessarcher we should be looking better here, let me know if you come across any other oddities.

I ended up moving the mb_wordwrap function to the `Truncates` trait, felt like it might be right?

I also moved the `$rows` argument to the last position so that the signature basically matched the signature of `text`, let me know if that's cool.

jessarcher:
Looks great @joetannenbaum! I made a few formatting tweaks to the code and also added support for the placeholder to wrap and/or contain newlines.

jessarcher (approved):

taylorotwell:
Thanks!

PR #97, opened 2023-10-21
Displaying a final message when rendering spinner

joetannenbaum:
This PR introduces an optional third argument to the `spin` helper, allowing the user to display a final message after the spinner has completed its process.

The message can either be a string:

```php
spin(
    function () {
        sleep(4);

        return 'A-OK';
    },
    'Checking system...',
    'System looks good!',
);
```

or a closure that receives the result as an argument:

```php
spin(
    function () {
        sleep(4);

        return '8.2';
    },
    'Detecting PHP Version...',
    fn ($result) => "PHP Version: <info>{$result}</info>",
);
```

Here's what it looks like:

https://github.com/laravel/prompts/assets/2702148/a1f8a719-5c47-4ef1-9d2d-95b01e36adc3

### DIFF
--- a/playground/spin.php
+++ b/playground/spin.php
@@ -2,7 +2,7 @@
 
 use function Laravel\Prompts\spin;
 
-require __DIR__ . '/../vendor/autoload.php';
+require __DIR__.'/../vendor/autoload.php';
 
 $result1 = spin(
     function () {
--- a/playground/spin.php
+++ b/playground/spin.php
@@ -2,9 +2,9 @@
 
 use function Laravel\Prompts\spin;
 
-require __DIR__.'/../vendor/autoload.php';
+require __DIR__ . '/../vendor/autoload.php';
 
-$result = spin(
+$result1 = spin(
     function () {
         sleep(4);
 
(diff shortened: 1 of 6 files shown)


joetannenbaum:
I realized yesterday after submitting this PR that since the spinner erases itself, you could just output a message yourself using something like `info` afterwards. 

I still like the idea of the spinner taking care of this for you, but if you're not into the idea of adding an additional argument let me know.

asciito:
> I realized yesterday after submitting this PR that since the spinner erases itself, you could just output a message yourself using something like `info` afterwards.
> 
> I still like the idea of the spinner taking care of this for you, but if you're not into the idea of adding an additional argument let me know.

Personally I prefer the addition of this feature, just because more than once I have to print a message just right before the spinner ends. So for example, if need to perform a lot of tasks and then print a message for every task, well, I really prefer this approach instead of writing more `info` functions every time.

mho22:
@joetannenbaum I created a quick artisan command to test the solution you proposed in case this pull request isn't merged, and it seems to have a minor newline issue : 

`app\Console\Commands\Test.php`

```
<?php

namespace App\Console\Commands;

use Illuminate\Console\Command;

use function Laravel\Prompts\spin;
use function Laravel\Prompts\info;


class Test extends Command
{
    protected $signature = "test";

    public function handle() : void
    {
        spin( fn() => sleep( 2 ), "Sleeping..." );

        info( "Awake" );

        spin( fn() => sleep( 2 ), "Sleeping..." );

        info( "Awake" );

        spin( fn() => sleep( 2 ), "Sleeping..." );

        info( "Awake" );
    }
}
```


https://github.com/laravel/prompts/assets/17177411/368b9b52-22c4-4591-a0b2-797d6500cd92



It appears that when a spin is called for the second time and an info is added afterward, there is a new line between them.

@jessarcher I suppose I should maybe add this as an issue ?

PR #108, opened 2024-02-08
Added pause prompt

allanmcarvalho:
This PR introduce the pause prompt that will be useful in a lot of scenarios. Eg.:

- User need to read and accept a license. Pressing ENTER he agrees with license, or press CTRL-C to cancel (sample bellow).
- The system need to know if users wants to continue with some action that can't be undone.
- The system need to make a pause and show some content for user before continue.

Captura de Tela 2024-02-08 às 14 19 52

### DIFF
--- a/playground/pause.php
+++ b/playground/pause.php
@@ -1,0 +1,38 @@
+<?php
+
+use Laravel\Prompts\Prompt;
+use function Laravel\Prompts\pause;
+
+require __DIR__.'/../vendor/autoload.php';
+
+Prompt::fallbackWhen(true);
+
+$license = "Copyright (c) 2023 Laravel LCC
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the \"Software\"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED \"AS IS\", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


jessarcher:
Hey @allanmcarvalho, I'm in favour of a `pause` function, but I'm not sure about the UI.

I've had a similar use case previously, but I just wanted a simple "Press enter to continue..." to allow the user to read the previous output. In that scenario, I wouldn't want a box with text, as the output I want them to read had already been written.

What are your thoughts on something more simple API like:

```php
function pause($message = 'Press enter to continue...'): void
{
    //...
}
```

Which would just output something similar to the `info` function:

image

In your case, you could output the license separately, followed by a `pause`. Although I'd probably use a `confirm` rather than `pause` to get an explicit yes/no.

allanmcarvalho:
Hi @jessarcher! Thanks by you review, you are definitely right. I will refactor with your suggestions ASAP and then push here. Have a good week!

allanmcarvalho:
@jessarcher I have made the suggestions that you provided.

```php
warning("Warning: The changes you are about to make cannot be undone. Please proceed with caution.");

pause();

info('Ok, processing...');
```

will output after press enter
Captura de Tela 2024-02-19 às 16 10 02

and then
Captura de Tela 2024-02-19 às 16 10 14

What do you think now?

jessarcher:
Hey @allanmcarvalho,

Looks good! I changed it so that when the terminal is non-interactive, the `pause` has no effect instead of throwing a validation error. The return value will be `false` if the prompt wasn't rendered in case someone needs to know that. I don't think `pause` should be used to get a confirmation for destructive actions - just to pause the output to allow the user to read the previous output before continuing execution. It's almost like a confirm prompt where you can't select "No" (but can obviously Ctrl-C, potentially leaving things only halfway complete).

As mentioned previously, for scenarios where you need explicit confirmation from the user before doing something potentially destructive, I'd recommend using the `confirm` prompt (along with a custom `--force` option to handle non-interactive scenarios). This is what we do in the framework with the [`ConfirmableTrait`](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Console/ConfirmableTrait.php) (See [`MigrateCommand`](https://github.com/laravel/framework/blob/08bf276de0efc9af130e10cb93ede827715543c3/src/Illuminate/Database/Console/Migrations/MigrateCommand.php#L79-L81) for example usage)

What are your thoughts?

allanmcarvalho:
Yes, I think that's it. Your point of view appears to well describe the objective behind of creation of this prompt. To be honest, I had the idea to create this prompt in a scenario where I needed to make a pause to read the previous content, so, it's like you described.

For scenarios where the user must to agree if continues or exit, maybe in future an exclusive prompt that made this like a charm can be crafted, but definitely it's not a job for pause.

Thanks for you review and contribution.

jessarcher (approved):

Issue #107, opened 2024-02-02
Allow easy selection of output interface (stdin/stderr)


mortenscheel:
Scenario: You have a command that outputs raw data (like json or csv) to the console, and it uses laravel/prompts. You want to save the output to a file, so you run `php artisan some:command > path/to/file.json`.
Unfortunately the prompt is using stdout, so it gets sent to the file and you don't see anything in the console.

I figured out a workaround that makes the prompt use stderr:
```php
$output = $this->getOutput()->getOutput();
if ($output instanceof ConsoleOutputInterface) {
    Prompt::setOutput($output->getErrorOutput());
}
```

It works great, but it would be nice if there was an easier and more intuitive way of doing it. Maybe an optional argument that allows you to pass an OutputInterface, or a static `Prompt::stderr()` method. And maybe consider using stderr by default in the next major release, unless there are any drawbacks that I'm not aware of.

driesvints:
Heya, if you could send in a PR we could look at we can consider it maybe. Thanks for your suggestion.

PR #109, opened 2024-02-09
Allow using stderr for prompt output

mortenscheel:
Sending the prompt output to `stdout` (which is the default behavior) can be problematic in situations where the output of the command needs to be redirected to a file, or piped to another command.
- If the output is redirected to a file, e.g. `> data.json`, the prompt is sent directly to the file and won't be displayed in the console. The command will keep running, waiting for input, but the user won't know what's going on.
- If the output is piped to another command, e.g. `|  grep "foo"`, the prompt might also be invisible, or if the receiving command expects structured data like csv or json, it might cause an error.

The fix is easy. Just send the prompt output to `stderr` in stead. That's what [Symfony's QuestionHelper](https://github.com/symfony/console/blob/b95e86a6ed738fa91f3853a34ce6819ce065de8d/Helper/QuestionHelper.php#L54-L56) does by default. And according to [this excellent post](https://fideloper.com/laravel-symfony-console-commands-stderr) by Chris Fidao, it's actually a well established unix convention.

Like I mentioned in #107, I think the best solution would be to use stderr by default, but I realize that this could be considered a breaking change. So in stead I've just added a simple `Prompt::useStderr()` method that makes it easy to send all prompts to stderr.

I really wanted to add some tests that assert that actual output is sent to stdout, and prompts are send to stderr, but I honestly have no idea how to do that.

### DIFF
--- a/src/Prompt.php
+++ b/src/Prompt.php
@@ -210,6 +210,7 @@ abstract class Prompt
     public static function useStderr(): void
     {
         $output = static::output();
+
         if ($output instanceof ConsoleOutputInterface) {
             static::setOutput($output->getErrorOutput());
         }
--- a/src/Prompt.php
+++ b/src/Prompt.php
@@ -5,6 +5,7 @@ namespace Laravel\Prompts;
 use Closure;
 use Laravel\Prompts\Output\ConsoleOutput;
 use RuntimeException;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Throwable;
 
@@ -201,6 +202,17 @@ abstract class Prompt
     public static function validateUsing(Closure $callback): void
     {
         static::$validateUsing = $callback;
+    }
+
+    /**
+     * Use stderr for prompt output
+     */
+    public static function useStderr(): void
+    {
+        $output = static::output();
+        if ($output instanceof ConsoleOutputInterface) {
+            static::setOutput($output->getErrorOutput());
+        }
     }
 
     /**
--- a/src/Prompt.php
+++ b/src/Prompt.php
@@ -5,7 +5,6 @@ namespace Laravel\Prompts;
 use Closure;
 use Laravel\Prompts\Output\ConsoleOutput;
 use RuntimeException;
-use Symfony\Component\Console\Output\ConsoleOutputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Throwable;
 
@@ -92,7 +91,7 @@ abstract class Prompt
                 return $this->fallback();
             }
 
-            static::$interactive ??= stream_isatty(STDIN);
+            static::$interactive ??= stream_isatty(STDIN) && (stream_isatty(STDOUT) || stream_isatty(STDERR));
 
             if (! static::$interactive) {
                 return $this->default();
@@ -202,18 +201,6 @@ abstract class Prompt
     public static function validateUsing(Closure $callback): void
     {
         static::$validateUsing = $callback;
-    }
-
-    /**
-     * Use stderr for prompt output
-     */
-    public static function useStderr(): void
-    {
-        $output = static::output();
-
-        if ($output instanceof ConsoleOutputInterface) {
-            static::setOutput($output->getErrorOutput());
-        }
     }
 
     /**
--- a/src/Prompt.php
+++ b/src/Prompt.php
@@ -3,8 +3,10 @@
 namespace Laravel\Prompts;
 
 use Closure;
-use Laravel\Prompts\Output\ConsoleOutput;
+use Laravel\Prompts\Output\PromptWriter;
 use RuntimeException;
+use Symfony\Component\Console\Output\ConsoleOutput;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Throwable;
 
@@ -65,9 +67,9 @@ abstract class Prompt
     protected static ?Closure $validateUsing;
 
     /**
-     * The output instance.
+     * The writer instance.
      */
-    protected static OutputInterface $output;
+    protected static PromptWriter $writer;
 
     /**
      * The terminal instance.
@@ -102,7 +104,7 @@ abstract class Prompt
             try {
                 static::terminal()->setTty('-icanon -isig -echo');
             } catch (Throwable $e) {
-                static::output()->writeln("<comment>{$e->getMessage()}</comment>");
+                static::writer()->write("<comment>{$e->getMessage()}</comment>", true);
                 static::fallbackWhen(true);
 
                 return $this->fallback();
@@ -154,9 +156,7 @@ abstract class Prompt
      */
     protected function capturePreviousNewLines(): void
     {
-        $this->newLinesWritten = method_exists(static::output(), 'newLinesWritten')
-            ? static::output()->newLinesWritten()
-            : 1;
+        $this->newLinesWritten = static::writer()->newLinesWritten();
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/Prompt.php
```diff
@@ -203,6 +204,17 @@ public static function validateUsing(Closure $callback): void
         static::$validateUsing = $callback;
     }
 
+    /**
+     * Use stderr for prompt output
+     */
+    public static function useStderr(): void
+    {
+        $output = static::output();
```
driesvints, line 212:
```suggestion
        $output = static::output();

```

taylorotwell:
I don't think we should have a method for this. The best practice should be the default.

taylorotwell:
Waiting for review from @jessarcher

jessarcher:
> I don't think we should have a method for this. The best practice should be the default.

Agreed. Based on the linked post, I think I can agree that using `STDERR` makes sense for interactive prompts. I'm less certain about where output from functions like `info` should go. Am I right in assuming that Symfony still uses `STDOUT` for their equivalent methods, and by extension, so would Laravel's wrappers?

Any changes we make would need to consider the newline behaviour. We output a single blank line between prompts and other informational messages so we need to track the number of trailing newlines previously output so we know how many (if any) need to be output before the next prompt. If output is mixed between `STDOUT` and `STDERR`, we'd still need to track the newlines globally between them, but I could see this causing some layout problems if `STDOUT` was being redirected somewhere else.

mortenscheel:
> Am I right in assuming that Symfony still uses STDOUT for their equivalent methods, and by extension, so would Laravel's wrappers?

Symfony commands don't have convenience methods like `info()` or `error()`. You just receive an `OutputInterface` like this
```php
protected function execute(InputInterface $input, OutputInterface $output): int
{
    $output->writeln('<info>This goes to STDOUT</info>');
    // If you want to output to STDERR, you have to do it manually
    $err = $output instanceof ConsoleOutputInterface ? $output->getErrorOutput() : $output;
    $err->writeln('<error>This goes to STDERR</error>');
    
    return self::SUCCESS;
}
```

Good point about `info()`. I personally think that prompt functions that can be considered "output formatter helpers" (info, note, error, table, etc.) should still use STDOUT, because they're used deliberately to produce output. Those that only produce temporary output (spin and table), as well as those that involve user interaction, should use STDERR. You could of course argue that `error()` messages should go to STDERR, but Artisan uses STDOUT, so it's probably best to stick with that.

> If output is mixed between STDOUT and STDERR, we'd still need to track the newlines globally between them, but I could see this causing some layout problems if STDOUT was being redirected somewhere else.

We should be able to detect if STDOUT is being redirected by using [this technique](https://github.com/sebastianbergmann/environment/blob/100d8b855d7180f79f9a9a5c483f2d960581c3ea/src/Console.php#L97-L122).

jessarcher:
> Good point about `info()`. I personally think that prompt functions that can be considered "output formatter helpers" (info, note, error, table, etc.) should still use STDOUT, because they're used deliberately to produce output. Those that only produce temporary output (spin and table), as well as those that involve user interaction, should use STDERR. You could of course argue that `error()` messages should go to STDERR, but Artisan uses STDOUT, so it's probably best to stick with that.

Makes sense, although it could be a pain if you want the command to output well-formed JSON that can be redirected to STDOUT, but still want to show some informational messages that don't get redirected.

> We should be able to detect if STDOUT is being redirected by using [this technique](https://github.com/sebastianbergmann/environment/blob/100d8b855d7180f79f9a9a5c483f2d960581c3ea/src/Console.php#L97-L122).

Currently, we only check whether STDIN is a TTY (or whether the `--non-interactive` option was passed to a Laravel command) to determine whether to prompt interactively.

What do you think about the following additional checks?

* If STDOUT is a TTY, then continue as usual.
* Otherwise, if STDERR is a TTY, use that for interactive output (or perhaps all output).
* If neither is a TTY, then disable prompt interactivity.

That should make things "just work" as expected for most scenarios.

mortenscheel:
> * If STDOUT is a TTY, then continue as usual.
> * Otherwise, if STDERR is a TTY, use that for interactive output (or perhaps all output). 
> * If neither is a TTY, then disable prompt interactivity.

Excellent solution 👍 I think I'll have time to implement it this week.

mortenscheel:
Looking forward to hearing your feedback.
Here's an easy way to test it:
```shell
php playground/index.php > out.txt
```
The prompt should work as always, and out.txt should only contain the final var_dump() with the selected values.

mortenscheel:
> Rather than duplicate the logic in Laravel, can we move it within the Prompt class so it works regardless of the OutputInterface being used?

Sure. Just to make sure I understand correctly, you're suggesting that the different prompts should call `static::write()` (where the counting happens before sending the message to an `OutputInterface`) in stead of `static::output()->write()`?

mortenscheel:
@jessarcher please take a look at the implementation.
I've refactored the newline capturing to a new `PromptWriter` class, which handles the capturing regardless of which `OutputInterface` is being used.

I was a bit confused by the fact that `Prompt` sometimes rendered without "synchronizing" the number of captured newlines from `ConsoleOutput`. So when I tried using a single `newLinesWritten` for everything, the prompt started jumping around quite a bit. I just wanted to mention it, in case it isn't intentional. Let me know if you want an example.

jessarcher:
Hi @mortenscheel,

I've just tested this out, and there's an issue with the newlines between Prompt output and Laravel output.

On the left is the current behaviour, and on the right is with this PR. Note the extra newlines between the Prompts output and the other output from Laravel. This is without redirecting any streams.

image

I'll need to do more testing to figure out what's going on, but I suspect it's to do with the two places where the `$newLinesWritten` value will be captured. In the `PromptWriter` class and in Laravel's `OutputStyle` class. Previously, it would only be captured by whichever output was being used, `ConsoleOutput` _or_ `OutputStyle` (but never both).

src/Prompt.php
```diff
@@ -154,37 +156,30 @@ public function newLinesWritten(): int
...
     /**
      * Set the output instance.
      */
     public static function setOutput(OutputInterface $output): void
     {
-        self::$output = $output;
+        if ($output instanceof ConsoleOutputInterface && stream_isatty(STDERR) && ! stream_isatty(STDOUT)) {
+            $output = $output->getErrorOutput();
+        }
+        self::$writer = new PromptWriter($output);
```
jessarcher, line 170: Laravel sends an instance of `\Illuminate\Console\OutputStyle`, which doesn't implement `ConsoleOutputInterface`, meaning we won't use STDERR with Laravel even when STDOUT is redirected elsewhere.

Laravel's `OutputStyle` class is an extension of Symfony's `SymfonyStyle` and `OutputStyle` classes, which are a wrapper around an underlying `OutputInterface`. It doesn't expose the `getErrorOutput` method but instead exposes a `getErrorStyle` method. Unfortunately, this returns a new instance of `SymfonyStyle` rather than Laravel's `OutputStyle` class (due to their use of `self`).

Even if we could get it to return a new instance of Laravel's `OutputStyle`, the newline tracking would break because Laravel will still be using its original `OutputStyle` instance rather than the new instance we'd be using in Prompts.

It's a complicated problem because we need to keep track of the newlines between Prompts's and Laravel's output. Both Laravel and Prompts need to know how many trailing newlines were in the previous output, regardless of where that previous output came from, so that they each know how many newlines to emit before any new output. If Laravel isn't aware of the trailing newlines that Prompts has written, it won't be able to output the correct amount of leading newlines before any new output, and vice versa. It gets more complicated with Symfony because they have a separate instance for STDERR. If Laravel and Prompts use different instances, they won't be aware of each other's trailing newlines.

The only solutions I can think of are:

1. Add logic to Laravel to make it use an STDERR output instance when appropriate (and share that instance with Prompts). This could have unintended consequences in user commands that expect to output on STDOUT regardless of any redirection, so it's probably a breaking change. Alternatively,
2. Add a method to Laravel's `OutputStyle` class (or create a new wrapper class) that allows us to write to STDERR, while still tracking trailing newlines regardless of the stream. It would potentially need to be smart enough to track newlines separately when streams are redirected to different places because, in that scenario, the trailing newlines from the previous output of another stream shouldn't be considered. Prompts could then choose to output to STDERR and both Laravel and Prompts would still know how many leading newlines they needed to output.

Either way it's a big change covering two code bases. I still see the value in doing this, but I'm concerned about the effort and risk involved.

mortenscheel: Thanks for the detailed explanation. I can't think of any other solutions, so you're welcome to close this if you want.
I learned a lot about console output, and it was nice to meet you guys 😄

jessarcher: Thanks, @mortenscheel! Nice to meet you as well :slightly_smiling_face: 

Happy to reconsider this at some point if we can figure out a good solution. Having one blank line between output looks great, but it certainly adds a lot of challenges :sweat_smile:

PR #110, opened 2024-02-10
Adds Support for Steps

lukeraymonddowning:
Hey folks!

This PR adds support for steps, which allow users to group related prompts together and go back if they make a mistake in a prompt.

Screenshot 2024-02-10 at 23 12 20

## Motivation

I created this because I often make a typo or select the wrong item in a prompt. Currently the only way to fix that is to cancel the entire command and restart, which can be frustrating if there are many prompts, such as in the Laravel installer.

## Usage

To create steps, you can use the `steps` helper function. From there, more prompts can be chained using the `then` method, which can be called for each "step" in the process. Finally, calling `run` will execute each prompt in order. An array of all responses will be returned as the final result from `run`.

```php
[$project, $database, $confirm] = steps(fn() => text('What should we call your project?'), title: 'Project Info')
    ->then(fn ($project) => select("Which database would you like to use for {$project}?", ['MySQL', 'PostGreSQL', 'SQLite']))
    ->then(fn() => confirm('Confirm project creation?'))
    ->run();
```

Note that each closure passed to `then` receives the result returned from the previous step. To undo a step, the user may press `CTRL+U`, which will delete the response and re-present the previous prompt in the step chain.

There may be situations where you want to run custom logic when a user reverts a step. For example, if the user has selected a database and we create it, we should delete it again if the user reverts that step. To facilitate this, we can provide a `revert` argument to `then`.

```php
steps(fn() => text('What should we call your project?'), title: 'Project Info')
    ->then(
        fn ($project) => select("Which database would you like to use for {$project}?", ['MySQL', 'PostGreSQL', 'SQLite']),
        revert: fn ($database) => info("Deleting database {$database}"),
    )
    ->run();
```

## For discussion

- Currently, there is no way to prevent undoing a step. It might be nice for us to provide a way to define in code if a step is irreversible. 
- I have called this "Steps", but it could just as well be called "Chains" or "Group" or similar. If there is a better terminology that fits the bill, I'm all ears.
- For history and clarity, undoing a prompt currently still shows previous steps. However, it might make sense to erase previous answers and present the modal fresh again.
- I have chosen `CTRL+U` as the shortcut for this, which is technically "negative acknowledgement". I did this because it is unused and is "undo". I wanted to get `CTRL+Z` working, but I think there are possibly protections around that combination. If somebody knows more than me on that topic, please teach me 🙂

As always, thank you all so much for the incredible work on Prompts and Laravel.

Kind Regards,
Luke


JonPurvis:
This is awesome @lukeraymonddowning! 

I've been using Prompts a lot recently and actually plan on using it at work to rewrite one of our CLI tools. This update would prove rather useful.

I agree that it would be nice to prevent certain steps from being irreversible. Perhaps a red warning underneath the Prompt itself too, to let the user know that whatever they select for this particular prompt, can't be changed. I also agree that if you do undo a prompt, it should remove previous submissions from the output, otherwise it could look messy!

It would be a real _Tragedy_ if this didn't make it in to the package
(^ rather niche joke, but I felt like it had to be done 🙃)

JustSteveKing:
Love this @lukeraymonddowning - this will make building more complicated a lot easier 🔥

lukeraymonddowning:
I'm going to close this, because I have a much nicer implementation in the works 👍

PR #116, opened 2024-02-21
[Feature] Reactive multi-select options through closure, plus auto-validation

GoodM4ven:
Changes:
- Added the ability to provide multiselect options as a closure [reactivity]
    - Created an evaluater for the closer to be called wherever the options are needed
    - Added a test for the feature, then tested everything
- Made the multi-select trigger upon selection (space key) [auto-validation]
    - Exposed validate method for that to happen

Concerns:
- Should there be a config check for whether to do the validation instantly? Or why doesn't validation happens upon selection by default?
- Honestly, I can't seem to understand how would I trigger the option re-evaluation without triggering validation first!

Issues:
- Need to re-evaluate the index, or else movement wouldn't be normal.
- Need to trigger show changes without triggering validation, because re-evaluation currently works only when everything is valid.

Todo:
- Docs; with a good example for reactivity!

Use-case:
https://github.com/laravel/prompts/assets/121377476/bfce061c-30e3-4456-93c5-84d1ec68e77a

_Comments:
Dude! How many places should they have the project-name shown at in VSCodium?!!! lol_

### DIFF
--- a/src/MultiSelectPrompt.php
+++ b/src/MultiSelectPrompt.php
@@ -151,7 +151,7 @@ class MultiSelectPrompt extends Prompt
      * Returns options as an array; and re-evaluates them if they were in a closure
      * 
      * @param  array<int|string, string>|Collection<int|string, string>|Closure  $options
-     * @return array
+     * @return array<int|string, string>
      */
     public function eval(array|Collection|Closure $options): array
     {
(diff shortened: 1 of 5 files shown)


driesvints:
@GoodM4ven feel free to resend this once you're ready.

Issue #115, opened 2024-02-21
Feature request: section support


henzeb:
I would love to see some support on sections.

Symfony has implementation for that, but was never really used in Laravel. 

What it does is creating a segment in which you can add or delete lines while also having other segments which can be edited independently from each other.  

One of the ideas is to be able to clear out prompts when entered information, but we can do so many things with sections. We could do live updating of sections using Laravel Reverb ;-)

An example of what could be done: 

```php
section('foo', function(){
     // anything in here happens within the section
     info('hello');
     // .... creating progressbars, requesting information.
})->clear();
info('we are done');
```

Another example: 
```php
info('asking questions:', section: 'foo');
$bar = section('bar');

$name = text('your name?', section: 'bar');
$bar->clear();
$password = password('your password?', section: $bar);
$bar->replace(
  function(){
     // replaces the entire content. Imagine it does more output.
     info('Thanks!');
  }
)

info('done asking questions.', section: 'foo');
```

henzeb:
After tinkering with Prompt a bit, I don't think sections are really needed for the watch method after all, They can be separated Feature Requests. I will PR the watch prompt somewhere this weekend.

driesvints:
Thanks! Let's see how your PR goes.

henzeb:
apparently it is not wanted. Made a package for it. https://packagist.org/packages/henzeb/prompts-extras

PR #117, opened 2024-02-24
Watch feature

henzeb:
### What?
Introducing `watch`. A simple feature that mimics the basic features of the `watch` function found on Linux/Unix systems.

### Why?

I've developed a lot of command line tools, and I often needed to monitor output in the console. Most of the time with tables and such.

Using the `watch` functionality on Linux systems should do it, but this requires the entire framework to be loaded every time, which can put a strain on the server capacity. With this, we don't even need cache, but we can use the `once` function instead for things that shouldn't be cached, but still are needed.

### Example

In the playground directory, there is an example with a table updating with new records every second. Here to just demonstrate the simplicity.

````php
use function Laravel\Prompts\table;
use function Laravel\Prompts\watch;

watch(
   watch: function(){
       table(['header'], [random_int()]);
   },
   interval: 2 // intervals are in seconds
);
````

###Testing

Watch comes with a few testing tools. 

#### fakeTimes

By default it iterates just once, but you can tell how many iterations `watch` should execute.

````php
use Laravel\Prompts\Prompts;
use Laravel\Prompts\Watch;
use function Laravel\Prompts\watch;
use function Laravel\Prompts\info;

Prompts::fake();
Watch::fakeTimes(2);

watch(
    function(){
       info('hello world');
    }
);

Prompts::content(); // would contain hello world twice
````

#### assertSecondsSleptBetweenIntervals

Allows you to assert the total seconds slept between intervals. Allows users to test if they passed the correct value.

````php
use Laravel\Prompts\Prompts;
use Laravel\Prompts\Watch;
use function Laravel\Prompts\watch;
use function Laravel\Prompts\info;

Prompts::fake();
Watch::fakeTimes(2);

watch(
    function(){
       info('hello world');
    }
);

Watch::assertSecondsSleptBetweenIntervals(4); // would succeed because default interval is 2 and it iterated twice.
````

#### shouldNotFakeIntervalSleep

This is not really a thing users would most likely use, but I used it to allow actually test sleeping. I can remove it if you don't want us to actually test sleep functions.

### Considerations

Right now, Prompt is resetting the cursor position using the lines count. When a user presses enter, this messes up the output.

I added a new method `isFaked` to the FakesInputOutput trait which can be used to detect if Prompt is currently being faked. This might be helpful in the future if things like sections as described in #115 are going to be added.

### DIFF
--- a/playground/watch.php
+++ b/playground/watch.php
@@ -1,0 +1,44 @@
+<?php
+
+use function Laravel\Prompts\info;
+use function Laravel\Prompts\note;
+use function Laravel\Prompts\progress;
+use function Laravel\Prompts\table;
+use function Laravel\Prompts\text;
+use function Laravel\Prompts\watch;
+
+require __DIR__ . '/../vendor/autoload.php';
+
+watch(
+    function () {
+        static $iteration = 0;
+        static $items = [];
+
+        if (count($items) === 5) {
+            array_shift($items);
+        }
+
+        $items[] = [$iteration += 1, (new Datetime())->format(DateTime::RFC850)];
+
+        if (count($items) === 5) {
+            info(sprintf('Now the table just scrolls, %s and counting...', $iteration));
+        } else {
+            info('Filling up the table...');
+        }
+
+        progress('a nice progressbar', 5)->advance($iteration % 5);
+
+        $ralph = text('Ralph', default: 'Ralph Wiggum: I\'m ignored!');
+
+        note($ralph);
+
+        table(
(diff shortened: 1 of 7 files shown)


taylorotwell:

If possible, please consider releasing your code as a package so that the community can still take advantage of your contributions! 

If you feel absolutely certain that this code corrects a bug in the framework, please "**@**" mention me in a follow-up comment with further explanation so that GitHub will send me a notification of your response.

PR #121, opened 2024-03-07
[Feature] Automatic validation option upon toggling multi-select prompts

GoodM4ven:
- Added an argument and a property "validateOnToggle" to MultiSelectPrompt that defaults to false
    - If set to true, `validated` property gets set to true upon `toggleHighlight`; which triggers validation
- Wrote a test for it
    - Ensured everything passes

I found this to be cool when it's accompanied with the next PR, which is about introducing re-evaluated and cached options closure; resulting in a reactive experience... Basically, it's what I was a bit confused about in #116 .

If you like it, should I do the docs?

### DIFF
--- a/src/MultiSelectPrompt.php
+++ b/src/MultiSelectPrompt.php
@@ -42,6 +42,7 @@ class MultiSelectPrompt extends Prompt
         public int $scroll = 5,
         public bool|string $required = false,
         public mixed $validate = null,
+        public bool $validateOnToggle = false, 
         public string $hint = '',
     ) {
         $this->options = $options instanceof Collection ? $options->all() : $options;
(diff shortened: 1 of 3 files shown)


jessarcher:
Hey @GoodM4ven, thanks for your PR!

I'm not sure it's worth adding an additional function parameter for this. I'd need to see some compelling use cases or multiple people experiencing an issue with the current behaviour.

Just to explain the reasoning behind the current behaviour, the delayed validation is intended to prevent an error message from being displayed before the user has had a chance to finish their input. This is primarily for text-based inputs where a user may be part-way through typing an email address. It also applies with multi-selects where the rule may be that they need to select a minimum number of options, and we don't want to display an error when they've chosen only one if they were already planning on selecting more.

PR #136, opened 2024-04-14
Feature: TabbedScrollableSelectPrompt

ProjektGopher:
This PR includes all work to implement a new `TabbedScrollableSelectPrompt`. 

https://github.com/laravel/prompts/assets/1688608/89f08a04-49bb-4d2d-bbcc-021f7cc3e91b

**If you'd like me to extract these into separate PRs, I can do that**

The first three commits are structured such that they can be `cherry-picked` and merged as separate work beforehand. They involve the following changes:
- Adds support for `ESCAPE`, `SHIFT_UP`, and `SHIFT_DOWN` Keys.
- Adds the ability to specify the character to be used in `DrawsBoxes::pad()`, maintaining backwards compatibility by setting a default of `' '`.
- Moves three methods for working with strings from the `DrawsBoxes` trait into another dedicated `HandlesStrings` trait which is then consumed by the `DrawsTabs` trait to avoid duplicate method name collisions.

The new `DrawsTabs` trait could then be merged on it's own. It's built in a similar way to the `DrawsScrollbars` trait. It can be used in any new prompts in the same way. This is the feature that ultimately enables the prompt in the example. It will also automatically scroll horizontally based on the selected tab in order to prevent overflowing and ruining the output.

Finally, the final 5 commits put all of this together to create our new component. I built this component because I wanted to include some readable context for the items from which I'm selecting. I'm thrilled with this solution, and I'd like to provide it for others to use as well.

### DIFF
--- a/playground/tabbedscrollableselect.php
+++ b/playground/tabbedscrollableselect.php
@@ -1,0 +1,129 @@
+<?php
+
+use function Laravel\Prompts\tabbedscrollableselect;
+
+require __DIR__.'/../vendor/autoload.php';
+
+$application = tabbedscrollableselect(
+    label: 'Which persons application would you like to choose?',
+    options: [
+        [
+            'id' => 0,
+            'tab' => 'Jess Archer',
+            'body' => <<<BODY
+            Subject: Application for Software Developer Position - The Code to Success!
+
+            Dear Hiring Manager,
+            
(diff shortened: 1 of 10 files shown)


edgrosvenor:
I think this would fit nicely in lots of places. The first one that comes to mind is the Laravel installer. Being able to add some context around choices like what auth scaffolding to use so we don't have to assume the user has already read the docs on all the options would make for a nice improvement.

Full disclosure, we are already using this (an early proof of concept version of it) in one of our apps and it makes for a really great user experience when asking someone to make a choice where a little more context than just a name is helpful.

ProjektGopher:
Closing for now to add modifications to existing code separately. Will rebase and resubmit afterwards.

Issue #133, opened 2024-04-14
Incorrect cursor position when text area contains multi-byte strings


macocci7:
### Laravel Prompts Version

0.1.18

### Laravel Version

11.3.0

### PHP Version

8.2.13

### Operating System & Version

Ubuntu 22.04 via WSL2

### Terminal Application

VS Code's integrated terminal

#### Issue

If there is a multibyte string in the previous line, the cursor moves downwards to the wrong position.
The same goes for moving up.

#### Findings

`$this->cursorPosition` in `Laravel\Prompts\TextareaPrompt` includes multibyte widths counted as 2 by `mb_strwidth()`, and I think it's necessary for determining cursor position in multiline text.
It seems the bug is caused by `$this->cursorPosition` passed in:
```php
        return $this->addCursor($this->wrappedValue(), $this->cursorPosition + $this->cursorOffset(), -1);
```
I'm posting here because I couldn't find a good solution.

### Steps To Reproduce

```php
textarea(
    label: 'Tell me a story.',
    default: "ａｂ\nｃｄｅｆ\nｇｈｉjklmnnopqrs\ntuvwxyz\n0123456789\n+-*/",
);
```
run this code and:
1. The last line will not be displayed.
2. Press [Home], press [End], then the last line will not be displayed.
3. Press [Home], press [↓], then the cursor will move to right of 'ｄ'.
4. Press [↓] after step 3 above, then the cursor will move to right of 'l'.
5. Press [↓] after step 4 above, then the cursor will move to right of '0'.
6. Press [↓] after step 5 above, then the cursor will move to right of '8'.

jessarcher:
@joetannenbaum are you able to take a look at this one please?

joetannenbaum:
Will do 👍

PR #137, opened 2024-04-17
Fix for up/down arrows + cursor position when textarea content contains multi-byte strings

joetannenbaum:
This PR is a fix for #133.

When the content of the textarea contains a multi-byte string, we need to _count the characters_ in the content, not _measure the width_ of the string to determine the cursor position. 

By using `count(mb_str_split($content))` instead of `mb_strwidth($content)`, we are able to more accurately place the cursor where it belongs whether or not the content contains multi-byte strings.

### DIFF
--- a/src/TextareaPrompt.php
+++ b/src/TextareaPrompt.php
@@ -113,7 +113,7 @@ class TextareaPrompt extends Prompt
         $lines = collect($this->lines());
 
         // Line length + 1 for the newline character
-        $lineLengths = $lines->map(fn ($line, $index) => mb_strwidth($line) + ($index === $lines->count() - 1 ? 0 : 1));
+        $lineLengths = $lines->map(fn ($line, $index) => count(mb_str_split($line)) + ($index === $lines->count() - 1 ? 0 : 1));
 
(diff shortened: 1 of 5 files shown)


src/TextareaPrompt.php
```diff
@@ -113,7 +113,7 @@ protected function handleUpKey(): void
         $lines = collect($this->lines());
 
         // Line length + 1 for the newline character
-        $lineLengths = $lines->map(fn ($line, $index) => mb_strwidth($line) + ($index === $lines->count() - 1 ? 0 : 1));
+        $lineLengths = $lines->map(fn ($line, $index) => count(mb_str_split($line)) + ($index === $lines->count() - 1 ? 0 : 1));
```
jessarcher, line 116: Any reason not to use `mb_strlen` instead?

joetannenbaum: Nope 🤦‍♂️

I thought about it but for some reason thought it would return the wrong value. Not... not sure why.

Updated, thank you!

jessarcher (approved):

PR #143, opened 2024-05-01
Add DrawsTabs trait with tests

ProjektGopher:
This PR adds a trait that can be used in the same way as the `DrawsScrollbars` or `DrawsBoxes` traits to draw a series of tabs in a prompt. The selected index will be highlighted, and the tabs scroll horizontally automatically to make sure that the highlighted tab is **always** fully visible.

Example Usage:
```php
$this->tabs(
    tabs: collect(['One', 'Two', 'Three', 'Four', 'Five', 'Six']),
    selected: 0,
    width: 60,
);
```

Example Output:
```sh
    ╭─────╮                                       
    │ One │  Two    Three    Four    Five    Six  
    ┴─────┴─────────────────────────────────────────────────────
 ```

This could potentially be used at the top of a `box` for tabbed navigation in order to give extra context. Here's an example of this trait being used for exactly that purpose in a component I've built:

https://github.com/laravel/prompts/assets/1688608/3da28aa7-d77d-4fb5-b58c-27b51fcdc900

### DIFF
--- a/src/Themes/Default/Concerns/DrawsTabs.php
+++ b/src/Themes/Default/Concerns/DrawsTabs.php
@@ -39,11 +39,11 @@ trait DrawsTabs
 
         // automatic horizontal tab scrolling
         if ($strippedWidth($top_row) > $width) {
+            $scroll = $selected / ($tabs->count() - 1);
             $chars_to_kill = $strippedWidth($top_row) - $width;
-            $percent = $selected / ($tabs->count() - 1);
-            $left = (int) round($percent * $chars_to_kill);
+            $offset = (int) round($scroll * $chars_to_kill);
             foreach ([&$top_row, &$middle_row, &$bottom_row] as &$row) {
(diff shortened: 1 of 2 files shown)


taylorotwell:
Not sure we have the bandwidth to take this on at the moment 😅 I wonder if there is a world where Prompts could maybe support stuff like this via packages / extensions?

ProjektGopher:
@taylorotwell technically with my three other PRs that you've merged this week, I could now package my trait and prompt separately and it would be compatible.

My ultimate goal here is to get this prompt of mine into the laravel installer so that we can provide some detailed context on our various front end stack options to newcomers directly within the installer.

If I were to package my prompt up instead, would you be okay with me adding a new dependency on it?

ProjektGopher:
I've released this trait (along with the prompt that it enables) in my `[Community Prompts Package](https://github.com/artisan-build/community-prompts)` so that others can use it. I'll leave this PR open though until it's either merged or closed by you folks since I'd still love it to become first party and available to _everyone_.

PR #146, opened 2024-05-10
fix(helper): ensure helpers can't be redeclared

NickSdot:
To make packages like `symfony/var-dumper` available globally you prepend the global Composer `autoload.php` like so:

```
// php.ini

auto_prepend_file = ${HOME}/.composer/vendor/autoload.php
```
image

From: https://symfony.com/doc/current/components/var_dumper.html

To reproduce, at least these must be installed.

```
// composer

laravel/installer (global)
symfony/var-dumper (global)
laravel/framework (project)
phpstan/phpstan (project)
```

Then run PhpStan in a project.

```
// cli

$ vendor/bin/phpstan analyse Common
```

Result.

```
Fatal error: Cannot redeclare Laravel\Prompts\text() (previously declared in ~/.composer/vendor/laravel/prompts/src/helpers.php:11) in XXX/vendor/laravel/prompts/src/helpers.php on line 13

Call Stack:
0.0325    1260304   1. {main}() XXX/vendor/bin/phpstan:0
0.0325    1261384   2. include('XXX/vendor/phpstan/phpstan/phpstan') XXX/vendor/bin/phpstan:119
0.0754    4122912   3. require('phar://XXX/vendor/phpstan/phpstan/phpstan.phar/bin/phpstan') XXX/vendor/phpstan/phpstan/phpstan:8
0.0754    4123296   4. _PHPStan_7961f7ae1\{closure:phar://XXX/vendor/phpstan/phpstan/phpstan.phar/bin/phpstan:13-125}() phar://XXX/vendor/phpstan/phpstan/phpstan.phar/bin/phpstan:125
0.1766   36372104   5. require_once('XXX/vendor/autoload.php') phar://XXX/vendor/phpstan/phpstan/phpstan.phar/bin/phpstan:66
0.1766   36378600   6. ComposerAutoloaderInit9c491b8531eec05ba41a11d9276a5749::getLoader() XXX/vendor/autoload.php:25
0.1838   39023408   7. {closure:XXX/vendor/composer/autoload_real.php:37-43}($fileIdentifier = '47e1160838b5e5a10346ac4084b58c23', $file = 'XXX/vendor/composer/../laravel/prompts/src/helpers.php') XXX/vendor/composer/autoload_real.php:45
```

### Notes
- I did a Composer global update, and installed phpstan in the same time. Didn't dig deeper what causes the bug. But same as with other Laravel `helper.php` functions, I think they should be protected against redeclaration. This PR adds the required checks.
- Tests pass. I don't think it will break something. However, give it a deep second thought please. :)
- Diff here on GH is messed. In PhpStorm it's better.

---

_Amazing_ work here btw, Jess. Thank you! 🫡❤️

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -5,203 +5,245 @@ namespace Laravel\Prompts;
 use Closure;
 use Illuminate\Support\Collection;
 
-/**
- * Prompt the user for text input.
- */
-function text(string $label, string $placeholder = '', string $default = '', bool|string $required = false, mixed $validate = null, string $hint = ''): string
-{
-    return (new TextPrompt(...func_get_args()))->prompt();
-}
-
-/**
- * Prompt the user for multiline text input.
- */
-function textarea(string $label, string $placeholder = '', string $default = '', bool|string $required = false, ?Closure $validate = null, string $hint = '', int $rows = 5): string
-{
-    return (new TextareaPrompt($label, $placeholder, $default, $required, $validate, $hint, $rows))->prompt();
-}
-
-/**
- * Prompt the user for input, hiding the value.
- */
-function password(string $label, string $placeholder = '', bool|string $required = false, mixed $validate = null, string $hint = ''): string
-{
-    return (new PasswordPrompt(...func_get_args()))->prompt();
-}
-
-/**
- * Prompt the user to select an option.
- *
- * @param  array<int|string, string>|Collection<int|string, string>  $options
- * @param  true|string  $required
- */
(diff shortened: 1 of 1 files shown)


jessarcher:
I think it's caused by the Laravel installer being installed globally. It depends on Prompts, which effectively makes it installed globally as well. With the `auto_prepend_file` setting in place, there are effectively two copies of Prompts installed.

Prompts didn't previously have the `function_exists` guard like other Laravel helpers because they are namespaced functions rather than global, but I don't see any issues adding it.

PR #147, opened 2024-05-10
Allow selecting all options in a `multiselect`

duncanmcclean:
Currently, when using the `multiselect` component, if you want to select all of the available options, you have to manually hit `space + down` until you've toggled all the options.

This is mostly fine. However, if you have 10 options, you might not want to force your users to manually select all the options. Instead, you might want to let them toggle a single option.

This PR adds a new `canSelectAll` property to the `MultiSelectPrompt` class, which if `true`, will show an "All" option at the top of the options:

CleanShot 2024-05-10 at 17 34 25

```php
multiselect(
    label: 'Select your favourite Laravel packages',
    options: [
        'laravel/prompts',
        'laravel/pulse',
        'laravel/horizon',
        'laravel/octane',
        'laravel/jetstream',
    ],
    canSelectAll: true,
);
```

When enabled, it'll select all the options and when disabled it'll deselect them all. In the case the user enables the "All" option but subsequently goes to select one of the options, the "All" option will be deselected.

An alternative UX for this could be to *somehow* disable all of the other options when the "All" option is toggled so they can't be changed. 

Let me know if you'd like me to make any changes and I'll do my best to do them. Otherwise, if this isn't something you want to add to Prompts, then that's fine as well, just an idea I had. ❤️


jessarcher:
Hey @duncanmcclean,

This is pretty cool!

My only real issue is the "All" option is included in the returned array, which could cause problems depending on what the user is doing with the response. We'd need to be careful filtering it out, though, in case someone already has their own "All" option which they expect to be returned.

I don't love that it gets injected into the `$options` array, but I can't think of a nicer way off the top of my head. I was thinking it should be a renderer concern, but the prompt class needs to know when it has been toggled.

I think the new parameter should also accept a string, allowing folks to change or localize the wording (similar to the `required` param). I'm also not totally sold on the parameter naming, but I don't have a better alternative now.

macocci7:
How about toggling when [Ctrl]+[A] is  pressed?

duncanmcclean:
> My only real issue is the "All" option is included in the returned array, which could cause problems depending on what the user is doing with the response. We'd need to be careful filtering it out, though, in case someone already has their own "All" option which they expect to be returned.
> I don't love that it gets injected into the $options array, but I can't think of a nicer way off the top of my head. I was thinking it should be a renderer concern, but the prompt class needs to know when it has been toggled.

Ah yeah, that's a problem.

I can probably refactor the "All" option to live in the renderer instead, rather than being merged into the actual options of the prompt.

> I think the new parameter should also accept a string, allowing folks to change or localize the wording (similar to the required param). I'm also not totally sold on the parameter naming, but I don't have a better alternative now.

Good idea 👍 

> How about toggling when [Ctrl]+[A] is pressed?

That's a good idea. In fact, I think I actually prefer it to what I've done in this PR in adding a separate option.

However, it seems like there's [already a keybinding for `ctrl + a`](https://github.com/laravel/prompts/blob/main/src/MultiSelectPrompt.php#L56) to navigate to the top of the options list so maybe we can't do that 🤔

duncanmcclean:
I've finally managed to carve out some time to work on refactoring this PR. 

I've renamed the property to `$selectAll` (still don't totally love it but will do for now) and allowed for a string to be passed through.

I've also started to dive into moving the "select all" logic into the renderer, as suggested. However, I'm struggling to figure out the best way to prepend an option to the list without actually merging it into `$this->options`. 🤔 

I'm likely missing something obvious, so if there's any pointers you can give me, that'd be great. 

Like I mentioned in my previous comment, I actually quite like (or might even prefer) it being a keyboard shortcut, rather than a separate visual option. It would actually be very simple to implement (I built it in 5 minutes as a POC 😆 ). However, it looks like `ctrl + a` (the most likely candidate) has already been taken.

jessarcher:
Keyboard shortcuts can be funny in the terminal. `Ctrl+A` is a [Readline/Emacs](https://en.wikipedia.org/wiki/GNU_Readline) thing (added in #70).

I believe it _technically_ only applies to typed input (i.e. horizontal movement), so I'm open to remapping it.

I'm thinking the following could be good:

* `Ctrl+A` selects all in `multiselect` and also when focused on the options list in `multisearch`. `Ctrl+E` would do nothing in these contexts.
* In a text input (including the text input of a `multisearch`) `Ctrl+A` and `Ctrl+E` move to the beginning and end of the input (i.e. unchanged)
* `Home` and `End` would remain unchanged everywhere (moving between the top and bottom of lists).

I'm hoping this is pretty straight-forward and would also remove the need for a parameter to be added, as this feature would always be active.

src/MultiSearchPrompt.php
```diff
@@ -43,15 +43,17 @@ public function __construct(
...
 
         $this->initializeScrolling(null);
 
         $this->on('key', fn ($key) => match ($key) {
             Key::UP, Key::UP_ARROW, Key::SHIFT_TAB => $this->highlightPrevious(count($this->matches), true),
             Key::DOWN, Key::DOWN_ARROW, Key::TAB => $this->highlightNext(count($this->matches), true),
-            Key::oneOf([Key::HOME, Key::CTRL_A], $key) => $this->highlighted !== null ? $this->highlight(0) : null,
-            Key::oneOf([Key::END, Key::CTRL_E], $key) => $this->highlighted !== null ? $this->highlight(count($this->matches()) - 1) : null,
+            Key::CTRL_A => $this->highlighted !== null ? $this->toggleAll() : null,
+            Key::CTRL_E => null,
```
duncanmcclean, line 54: This probably looks a bit silly...

It's the only way I could find of preventing Prompts from moving to the search input when the user presses `cmd + e` while browsing options. Maybe that's fine, maybe there's a better way?

src/MultiSearchPrompt.php
```diff
@@ -43,15 +43,17 @@ public function __construct(
...
+            Key::oneOf([Key::HOME], $key) => $this->highlighted !== null ? $this->highlight(0) : null,
+            Key::oneOf([Key::END], $key) => $this->highlighted !== null ? $this->highlight(count($this->matches()) - 1) : null,
```
duncanmcclean, line 56: This might be intentional but wanted to mention it in case it's not...

When I tried to simplify this to `Key::HOME => ...`, the keybinding wouldn't work at all and the "supports the home and end keys while navigating options" test would fail, which is why I've left this in `Key::oneOf()`.

duncanmcclean:
@jessarcher Thanks for your suggestions - I've updated this PR to use keyboard shortcuts, rather than the separate "All" option. 

This PR is ready for review, whenever you're able.

src/MultiSearchPrompt.php
jessarcher: There are multiple escape sequences for the `Home` and `End` keys for various terminals, so their constants are arrays rather than strings:

```php
const HOME = ["\e[1~", "\eOH", "\e[H", "\e[7~"];

const END = ["\e[4~", "\eOF", "\e[F", "\e[8~"];
```

src/MultiSearchPrompt.php
jessarcher: I think it's good. The match statement's `default` branch calls `$this->search()` so that all other key presses will update the search results, which resets the list position. The only way to prevent that for a specific key is to make sure the `default` branch isn't triggered.

src/MultiSearchPrompt.php
```diff
@@ -132,6 +134,20 @@ public function visible(): array
...
+    protected function toggleAll(): void
+    {
+        if (count($this->values) === count($this->matches)) {
+            $this->values = [];
+        } else {
+            $this->values = $this->isList()
+                ? array_combine(array_values($this->matches), array_values($this->matches))
+                : array_combine(array_keys($this->matches), array_values($this->matches));
+        }
+    }
```
jessarcher, line 149: I'm not totally sure about this behaviour.

When I was playing around with it, I tried the following:

1. Perform a search (e.g. search for "james" in the playground) and select one or more options.
2. Perform another search that excludes the previously selected options (e.g. search for "nuno") and then use `Ctrl+A` to select all of the options.

I expected that the previously selected option(s) (i.e. James) would still be selected (which you can see by clearing the search or submitting it) but it wasn't.

I also tried this:

1. Perform a search (e.g. search for "james" in the playground) and select an option.
2. Perform another search that returns a single option that is not the previously selected option (e.g. search for "nuno") and then select that option using the space bar.
3. Now press `Ctrl+A`.

Due to the count comparison with the currently selected values, it will think nothing is selected and will try to select Nuno (who is already selected) and silently unselect James.

Curious to hear your thoughts on how you'd expect it to behave in these scenarios.

I think it should probably determine whether all of the current matches are included in the values to determine whether it's a select or unselect operation. And I think it should always maintain any values that were previously selected that aren't included in the current matches regardless of whether it's selecting-all or unselecting-all.

jessarcher:
Thanks @duncanmcclean! I've made a few minor tweaks to the key ordering just to keep related things together in my head.

Just need to figure out the right approach for that `multisearch` behaviour. I can probably jump in if you need a hand.

src/MultiSearchPrompt.php
duncanmcclean: I see... that makes sense!

src/MultiSearchPrompt.php
duncanmcclean: > I think it should probably determine whether all of the current matches are included in the values to determine whether it's a select or unselect operation. And I think it should always maintain any values that were previously selected that aren't included in the current matches regardless of whether it's selecting-all or unselecting-all.

Yeah, I agree, it should probably be checking to see if all the current matches are included in the values, rather than checking the values/matches counts. 

I'll aim to get this PR updated today. 👍

duncanmcclean: I've just pushed up a commit which should hopefully sort the behaviour. Let me know what you think.

src/MultiSearchPrompt.php
```diff
@@ -132,6 +134,27 @@ public function visible(): array
...
+     */
+    protected function toggleAll(): void
+    {
+        $allMatchesSelected = collect($this->matches)->every(fn ($label, $key) => $this->isList()
+            ? array_key_exists($label, $this->values)
+            : array_key_exists($key, $this->values));
+
+        if ($allMatchesSelected) {
+            $this->values = array_filter($this->values, fn ($value) => $this->isList()
+                ? ! array_key_exists($value, $this->matches)
```
jessarcher, line 148: The "unselect all" behaviour doesn't seem to work when using a list. 

I _think_ this fixes it:

```suggestion
                ? ! in_array($value, $this->matches)
```

src/MultiSearchPrompt.php
jessarcher: Nice! The new behaviour feels really intuitive.

jessarcher:
(I've fixed the phpstan issue in the `main` branch)

src/MultiSearchPrompt.php
duncanmcclean: Whoops! That seems to fix it for me, have pushed it up in a commit. Thank you!

jessarcher (approved):
Thanks! Really dig this feature.

pelmered:
There is an example in the documentation for a `canSelectAll` option [here](https://laravel.com/docs/11.x/prompts#multiselect), but there is nothing in the code. When I googled I ended up here, but it seems like that option was reverted and replaced with a keyboard shortcut. So I guess that example should be deleted from docs then?

Great feature anyway!

PR #150, opened 2024-05-28
Feature: FileSelector

macocci7:
This PR adds support for file selector.

Here's a demo video, which you can play with for yourself using `php playground/fileselector.php` on this branch.

https://github.com/laravel/prompts/assets/19181121/03c6f46d-a19d-4de0-af71-66b65c33e499

### Purpose

`fileselector` supplies auto complete function in interactive mode.
[Creating a file selector-like function is possible](https://gist.github.com/macocci7/e78a3670faa4a8d7d023866db531cb40#file-selectfile-php) by using `suggest` prompt with callback for `options`.
However, automatically updating options after auto completion is impossible.
And filterling options with file  `extensions` might be vervose.
Using `fileselector` solves these issues and makes codes more simple.

### Usage

```php
use function Laravel\Prompts\fileselector;

fileselector(
    label: 'Select a file to import.',
    placeholder: 'E.g. ./vendor/autoload.php',
    validate: fn (string $value) => match (true) {
        !is_readable($value) => 'Cannot read the file.',
        default => null,
    },
    hint: 'Input the file path.',
    extensions: [
        '.json',
        '.php',
    ],
);
```
This function lists the entries in the directory on the local file system that matches the input, as the options for suggest.

You can automatically complete the input like the `suggest` prompt by pressing `TAB` on the selected option, then continue to input.
Press `Enter` to finish input.
Actually this function was created from the copy of the `suggest` promt.

You can filter the options with the parameter `extensions`.
If the parameter `extensions` is specified, the path and directories whose path ends match one of the extensions array elements are returned as options.

### Points to concern

- I don't really like the way of implementation for the option update process after input completion.(`FileSelector::autoComplete()`) Please let me know if there is a better way.
- I removed the `SHIFT_TAB` key binding in order to use `TAB` for input completion. Is there a problem?
- I am concerned that I was not able to implement the stub as a file list during testing.

### DIFF
--- a/playground/fileselector.php
+++ b/playground/fileselector.php
@@ -8,7 +8,7 @@ $model = fileselector(
     label: 'Select a file to import.',
     placeholder: 'E.g. ./vendor/autoload.php',
     validate: fn (string $value) => match (true) {
-        !is_readable($value) === 0 => 'Cannot read the file.',
+        !is_readable($value) => 'Cannot read the file.',
         default => null,
     },
     hint: 'Input the file path.',
--- a/playground/fileselector.php
+++ b/playground/fileselector.php
@@ -12,6 +12,11 @@ $model = fileselector(
         default => null,
     },
     hint: 'The model name should be singular.',
+    extensions: [
+        '.md',
+        '.json',
+        '.php',
+    ],
 );
 
 var_dump($model);
--- a/playground/fileselector.php
+++ b/playground/fileselector.php
@@ -5,15 +5,14 @@ use function Laravel\Prompts\fileselector;
 require __DIR__.'/../vendor/autoload.php';
 
 $model = fileselector(
-    label: 'What model should the policy apply to?',
-    placeholder: 'E.g. User',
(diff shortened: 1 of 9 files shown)


taylorotwell:

If possible, please consider releasing your code as a package so that the community can still take advantage of your contributions! 

If you feel absolutely certain that this code corrects a bug in the framework, please "**@**" mention me in a follow-up comment with further explanation so that GitHub will send me a notification of your response.

macocci7:
OK.
Thank you for your suggestion!
I've released the code as my package [file-selector-prompt](https://github.com/macocci7/file-selector-prompt).

ProjektGopher:
@macocci7 I'd love to add this into my [Community Prompts](https://github.com/artisan-build/community-prompts) package! I'll open an issue on your package as well, in case this comment doesn't get surfaced.

macocci7:
@ProjektGopher I opened PR [#5](https://github.com/artisan-build/community-prompts/pull/5) on Community Prompts.
Please check it out.

PR #152, opened 2024-05-30
Add the `Conditionable` trait to the `FormBuilder`

ash-jc-allen:
Hey! This PR proposes adding the `Illuminate\Support\Traits\Conditionable` trait to the `Laravel\Prompts\FormBuilder` class.

I love using `when` in my Laravel apps so I can chain my method calls together. So I reached out today to use it when building a form in Prompts and noticed it didn't exist.

Here's an example use case of a form using a standard `if` statement to add a new user to the app:

```php
// Imagine this is coming from config or something like that...
$smsEnabled = true;

$form = form()
    ->text(
        label: 'Email:',
        required: true,
        validate: ['email'],
        name: 'email'
    );

if ($smsEnabled) {
    $form->text(label: 'Phone number:', name: 'phone_number');
    $form->confirm(label: 'Send Welcome SMS?', name: 'send_sms');
}

$formData = $form->submit();
```

Using `when` it could be rewritten like so:

```php
// Imagine this is coming from config or something like that...
$smsEnabled = true;

$formData = form()
    ->text(
        label: 'Email:',
        required: true,
        validate: ['email'],
        name: 'email'
    )
    ->when(
        $smsEnabled,
        function (FormBuilder $form): void {
            $form->text(label: 'Phone number:', name: 'phone_number');
            $form->confirm(label: 'Send Welcome SMS?', name: 'send_sms');
        },
    )
    ->submit();
```

I did notice that Prompts isn't specific to Laravel and can be used in other PHP apps, so I don't know if keeping the dependencies lightweight will play a part in the decision to merge this. But from looking at the Composer files, I think Prompts includes `illuminate/collections` which in turn requires `illuminate/conditionable`. So if I'm right, it means we already have access to this trait without needing to add any new dependencies.

If this is something you think might be helpful to other devs, please give me a shout if you might want anything changed on this PR. If not, that's totally cool, at least I gave it a shot! 😄

### DIFF
--- a/src/FormBuilder.php
+++ b/src/FormBuilder.php
@@ -4,10 +4,13 @@ namespace Laravel\Prompts;
 
 use Closure;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Traits\Conditionable;
 use Laravel\Prompts\Exceptions\FormRevertedException;
 
 class FormBuilder
 {
+    use Conditionable;
+
     /**
      * Each step that should be executed.
      *
--- a/tests/Feature/FormTest.php
+++ b/tests/Feature/FormTest.php
@@ -1,5 +1,6 @@
 <?php
 
+use Laravel\Prompts\FormBuilder;
 use Laravel\Prompts\Key;
 use Laravel\Prompts\Prompt;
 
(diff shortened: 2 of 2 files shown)


driesvints:
Sorry @ash-jc-allen but I actually want to advocate against this change as we're actually looking at removing the `illuminate/support` dependency entirely in the near future. It's now a circular dependency with framework core and we want to get rit of this. Adding functionality now which we'll have to remove later on doesn't help 😞

ash-jc-allen:
Hey @driesvints! That's totally cool, I had a feeling the dependencies might have been a bit of a concern 🙂

driesvints:
Cool. Gonna close this then.

PR #154, opened 2024-06-11
Extract Looping Mechanisms

ProjektGopher:
I've successfully implemented an asynchronous, non-blocking prompt, which can simply extend our existing abstract prompt, leaving backwards compatibility unaffected. By using a `reactphp` event loop, I'm able to trigger renders and do other operations at any time since reading terminal input is no longer a blocking operation. This can almost certainly remove our reliance on `forking` using the `Process Control` extension in the `Spinner` class, ~~and there's a **slight** chance that it could solve our Windows story (I still have to verify this, so don't quote me)~~. It also happens to enable me to do some **really cool stuff** involving _websockets_ that I'll have ready soon.

However, **none** of that is in _this_ PR.

This PR is exclusively to make my new feature compatible with Laravel Prompts irrespective of its ultimate home (I'm totally fine with it living in my `community prompts` package if that's what has to happen).

- I've updated any references to `self` to `static` when setting or getting the `ConsoleOutput` object so that the `$output` static property is set on the **child**.
- I've extracted the loop for registering keypress mocks into a method that can be overridden in an extending class.
- I've extracted the looping mechanism in the `prompt` method which can now _also_ be overridden in an extending class.
- I've added an empty `Nothing` support class.

Since the `while` loop is now a function, if we were to rely on returning `null` to continue looping, we could no longer return `null` from any prompt. Since `null` is a valid return value, I needed a way to, in effect, check for a `void` return.

**edit**: I checked, and non-blocking i/o still wont work on Windows.

### DIFF
--- a/src/Concerns/FakesInputOutput.php
+++ b/src/Concerns/FakesInputOutput.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Prompts\Concerns;
 
+use Closure;
 use Laravel\Prompts\Output\BufferedConsoleOutput;
 use Laravel\Prompts\Terminal;
 use PHPUnit\Framework\Assert;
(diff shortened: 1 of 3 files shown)


ProjektGopher:
Based on the following comment from @jessarcher in #77, I tested making the `text` prompt async and using it in a new command in a fresh project to confirm that the output buffers will work as expect. Everything appears fine.

> When Prompts are used with Laravel commands, we set the output buffer to the existing one we already have in the Laravel command to track newlines between Prompt output and other output in the command. The ConsoleOutput class is only used when using Prompts without Laravel.

taylorotwell:
Drafting pending @jessarcher taking a look.

taylorotwell:
After chatting with @jessarcher I think we're going to hold off on this one!

PR #160, opened 2024-09-06
Adding "Clear" Function For Cleaning The Terminal

TarsisioXavier:
## The Feature
This is a new function that can be used to clear the user's terminal.  
It doesn't break anything else since it works with it's own classes.  

## Possibly Usage
- Cleaning the terminal before running the script for a better user experience.  
- Erase sensitive information outputted by the script or inputted by the user.

### DIFF
--- a/playground/clear.php
+++ b/playground/clear.php
@@ -1,37 +1,19 @@
 <?php
 
 use function Laravel\Prompts\clear;
-use function Laravel\Prompts\info;
 use function Laravel\Prompts\note;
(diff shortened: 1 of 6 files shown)


jessarcher:
Thanks @TarsisioXavier

I made a few small tweaks, mostly so that any output after `clear` won't have top padding.

| Before | After |
|-|-|
| image | image |

jessarcher (approved):

PR #161, opened 2024-09-09
Adding Facade For Calling The Functions

TarsisioXavier:
## The Feature 📦  
In Laravel facades are a common thing to have, so I've made this small class which calls the functions of the helper file via static calls.  

This PR adds a convenient way to call the functions without importing every function the dev uses in his script. I’ve also added a way to inject this facade class as an object - if the dev desires - and call the functions using “->”.  

## Possible Usages 📝  
- Can be injected in a class or method like a controller for example.  
- Doesn't need to import all functions used in the script.  

| Before | After |
|---|---|
| Before | After |

### DIFF
--- a/playground/facade/clear.php
+++ b/playground/facade/clear.php
@@ -1,0 +1,17 @@
+<?php
+
+use Laravel\Prompts\Support\Facades\Prompt;
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+Prompt::note('This will disappear.');
+
+Prompt::pause('Press [Enter] to continue.');
+
+Prompt::clear();
+
+Prompt::note('This will also disappear.');
+
+Prompt::pause('Press [Enter] to continue.');
+
+Prompt::clear();
--- a/playground/facade/confirm.php
+++ b/playground/facade/confirm.php
@@ -1,0 +1,18 @@
+<?php
+
+use Laravel\Prompts\Support\Facades\Prompt;
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+$confirmed = Prompt::confirm(
+    label: 'Would you like to install dependencies?',
+    validate: fn($value) => match ($value) {
+        false => 'You must install dependencies.',
+        default => null,
+    },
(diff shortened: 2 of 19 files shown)


devajmeireles:
Genius!

devajmeireles:
Genius! But wouldn't this PR be aimed at adding only the Facade? I mean, maybe you don't need things as examples, after all it's the same as the same.

jessarcher:
Hey @TarsisioXavier,

If you want to reduce the number of imports, you could import the `Laravel\Prompts` namespace and reference the functions as follows:

```
<?php

use Laravel\Prompts;

$name = Prompts\text('What is your name?');
```

Also, I'd avoid calling the class you've created a facade because it doesn't extend Laravel's `Facade` class so it doesn't have the facade features.

devajmeireles:
> Hey @TarsisioXavier,
> 
> If you want to reduce the number of imports, you could import the `Laravel\Prompts` namespace and reference the functions as follows:
> 
> ```
> <?php
> 
> use Laravel\Prompts;
> 
> $name = Prompts\text('What is your name?');
> ```
> 
> Also, I'd avoid calling the class you've created a facade because it doesn't extend Laravel's `Facade` class so it doesn't have the facade features that the name implies.

Do you agree with the idea to have a Facade for this?

TarsisioXavier:
> Genius! But wouldn't this PR be aimed at adding only the Facade? I mean, maybe you don't need things as examples, after all it's the same as the same.

Thanks for the compliment @devajmeireles. Yeah those files wore only for me to verify if everything behave the same way, I only wanted to be sure, so they are completely disposable.

TarsisioXavier:
Good morning @jessarcher,

I did notice that you guys didn't make any facade and I was wondering why, so I tried my best to create one without the `illuminate/support` dependency. 😅  
Even though it doesn't extend from Laravel facade class it follows the design pattern of a facade forwarding the function call to the "real" function. 🙂
Since there is no intention to add a facade for this package I guess we can close this PR then? 🥲

PR #162, opened 2024-09-10
Extract Looping Mechanisms

ProjektGopher:
This is a resurrection of #154 but with some stronger typing, and removing some breaking changes that were pointed out by @jessarcher to me in Dallas.

This PR should now include no behavioural changes, meaning that the existing test suite should cover it.

The limitation this PR is solving is making the looping mechanisms in the `prompt` method, and in our `fake` method overridable. This will allow us to use novel approaches that might unlock some other really cool possibilities, like debounced http requests, handling streamed output, and websockets.


taylorotwell:
Thanks @ProjektGopher!

PR #171, opened 2024-09-25
Adds support for conditional steps

jochem-blok:
In pull request #118 revertable forms is introduced. I found no proper way to make steps conditional to the answer of the previous steps. The `$step` closure seems to be an option. But then reverting does not work.

See the example below:
```php
use function Laravel\Prompts\form;

$responses = form()
    ->text('What is your name?')
    ->select('What is your language?', ['PHP', 'JS'], name: 'language')
    ->add(function ($responses) {
        if ($responses['language'] == 'PHP') {
            return text("Which version?");
        }
    },
        name: 'phpversion',
    )
    ->add(function ($responses) {
        if ($responses['language'] == 'JS') {
            return text("Which os?");
        }
    },
        name: 'os'
    )->confirm('Are you sure?')
    ->submit();
```

afbeelding

The option `ignoreWhenReverting` seems to be an option, but it that simply skips the step. So conditional forms are hard.

In the example below I rewrote the above example with the suggested change. And now everything seems to work fine.

```php
use function Laravel\Prompts\form;

$responses = form()
    ->text('What is your name?')
    ->select('What is your language?', ['PHP', 'JS'], name: 'language')
    ->add(function () {
            return text("Which version?");
        },
        name: 'phpversion',
        condition: function ($responses) {
            return $responses['language'] == 'PHP';
        }
    )
    ->add(function () {
            return text("Which os?");
        },
        name: 'os',
        condition: function ($responses) {
            return $responses['language'] == 'JS';
        }
    )->confirm('Are you sure?')
    ->submit();
```

afbeelding

### DIFF
--- a/src/FormBuilder.php
+++ b/src/FormBuilder.php
@@ -25,7 +25,17 @@ class FormBuilder
     /**
      * Add a new step.
      */
-    public function add(Closure $step, ?string $name = null, bool $ignoreWhenReverting = false, bool|Closure $condition = true): self
+    public function add(Closure $step, ?string $name = null, bool $ignoreWhenReverting = false): self
+    {
+        $this->steps[] = new FormStep($step, true, $name, $ignoreWhenReverting);
+
+        return $this;
+    }
+
+    /**
+     * Add a new conditional step.
+     */
+    public function addIf(Closure $condition, Closure $step, ?string $name = null, bool $ignoreWhenReverting = false): self
     {
         $this->steps[] = new FormStep($step, $condition, $name, $ignoreWhenReverting);
 
--- a/src/FormBuilder.php
+++ b/src/FormBuilder.php
@@ -25,9 +25,9 @@ class FormBuilder
     /**
      * Add a new step.
      */
-    public function add(Closure $step, ?string $name = null, bool $ignoreWhenReverting = false): self
+    public function add(Closure $step, ?string $name = null, bool $ignoreWhenReverting = false, bool|Closure $condition = true): self
     {
-        $this->steps[] = new FormStep($step, true, $name, $ignoreWhenReverting);
+        $this->steps[] = new FormStep($step, $condition, $name, $ignoreWhenReverting);
 
         return $this;
     }
--- a/src/FormBuilder.php
+++ b/src/FormBuilder.php
@@ -35,7 +35,7 @@ class FormBuilder
     /**
      * Add a new conditional step.
      */
-    public function addIf(Closure $condition, Closure $step, ?string $name = null, bool $ignoreWhenReverting = false): self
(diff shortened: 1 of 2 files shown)


taylorotwell:
Drafting pending review from @jessarcher

jessarcher (approved):

taylorotwell:
Should this instead be something like `addIf` where the condition can be the first argument?

jochem-blok:
Is this something I should answer? This is my first PR :)

I am happy to implement the suggested change.

devajmeireles:
> Is this something I should answer? This is my first PR :)
> 
> I am happy to implement the suggested change.

He's asking a general question, to all PR participants (including you), and also to the main maintainer, @jessarcher

jessarcher:
Having the condition as the first argument reads nicely.

I think the ultimate API would be a `when` method that gets the form instance:

```php
$responses = form()
    ->text('First question...')
    ->when($condition, fn ($form) => $form
        ->text('Conditional question')
        ->text('Another conditional question')
    );
```

But I had a quick play around and it's not straight-forward to implement with the revert behavior.

jochem-blok:
Following your reasoning Taylor why isn't addScrollable (because scroll is also an option) added? Having each optional parameter in a function would create a cartesian product.

How do we advance in this? Jess is trying to implement the same effect with another solution.

jessarcher:
> Following your reasoning Taylor why isn't addScrollable (because scroll is also an option) added? Having each optional parameter in a function would create a cartesian product.

I'd say it's because the conditional aspect, when present, is important to read _before_ the specific configuration of the step. It follows the way a typical `if` statement reads (_if, then..._) and matches things like the `abort_if` function in the framework.

> How do we advance in this? Jess is trying to implement the same effect with another solution.

I think Taylor would be happy to merge if you added a new `addIf($condition, ...)` method, instead of changing the `add` method.

jochem-blok:
I implemented the new method and changed the test cases. Please review :)

src/FormBuilder.php
```diff
@@ -32,6 +32,16 @@ public function add(Closure $step, ?string $name = null, bool $ignoreWhenReverti
         return $this;
     }
 
+    /**
+     * Add a new conditional step.
+     */
+    public function addIf(Closure $condition, Closure $step, ?string $name = null, bool $ignoreWhenReverting = false): self
```
jessarcher, line 38:
```suggestion
    public function addIf(bool|Closure $condition, Closure $step, ?string $name = null, bool $ignoreWhenReverting = false): self
```

jessarcher (approved):

src/FormBuilder.php
jochem-blok: What is the purpose of having a step with a false condition? Why would you add such a step?

PR #186, opened 2025-03-08
Allow static validator to run before prompt's validator.

AlexSkrypnyk:
## Summary

Prompts allow to specify a validator for a prompt using `validate()` closure or static `$validateUsing` closure.

`Prompt::validate()` contains a `match()` with:
```php
$error = match (true) {
    is_callable($this->validate) => ($this->validate)($value),
    isset(static::$validateUsing) => (static::$validateUsing)($this),
    default => throw new RuntimeException('The validation logic is missing.'),
};
```

This will call the Prompt's validator before the static validator.

## Problem

When testing prompts (via PHPUnit or Pest) using interactive inputs, the prompt that has failed the validation will halt the test execution expecting an input.

To work around this, a test executor (PHPUnit in the example below), could override the handling of the incorrect validator to, for example, throw an exception instead of re-asking for an input.

```php
// Fake input.
Prompt::fake(['t', 'e', 's', 't', Key::ENTER]);

// Intercept handling of the validation by throwing an exception.
Prompt::validateUsing(function (Prompt $prompt) {
    if (is_callable($prompt->validate)) {
        $error = ($prompt->validate)($prompt->value()); // <--- similar to what Prompt::validate() does.
        
        // Intercepting and throwing an exception that will later be picked up by a test and would fail an assertion.
        if ($error) {
            throw new \RuntimeException(sprintf('Validation for "%s" failed with error "%s".', $prompt->label, $error));
        }
    }

    return NULL;
  });
  
// call the prompt  
```

Before this PR, the static `$validateUsing` will not be assessed if the Prompt's `validate()` closure is set, making it impossible to intercept the validation handling.

After this PR is merged, the static `$validateUsing` will take precedence and will allow to intercept the validation.


src/Prompt.php
```diff
@@ -249,7 +249,7 @@ public static function terminal(): Terminal
     /**
      * Set the custom validation callback.
      */
-    public static function validateUsing(Closure $callback): void
+    public static function validateUsing(?Closure $callback): void
```
AlexSkrypnyk, line 252: Note that `$validateUsing` property [already allows](https://github.com/laravel/prompts/pull/186/files#diff-669bd09458cf69d12a4c7ddae353f28eb26cf6bb4ee3d94fd5fa4aabb65b2450R77) to be set to `NULL`.

AlexSkrypnyk:
@jessarcher 
Thank you for reviewing the PR.

I've looked through the functionality you are referring to in [laravel](https://github.com/laravel/framework/pull/49497/files#diff-9c0b99959f28e3b7a54a93a4ff67818de374a378726f8ecd16fa3c3bded6a9c7R32) and can see that the [`validatePrompt()`](https://github.com/laravel/framework/pull/49497/files#diff-9c0b99959f28e3b7a54a93a4ff67818de374a378726f8ecd16fa3c3bded6a9c7R169) expects rules to be passed as an array.

I agree that this change will break that functionality if `validate` is callable.

But from the point of Prompts, Laravel is a consumer project and should be able to adapt to changes in Prompts.

`validatePrompt()` in Laravel can be updated to either evaluate the `validate` callback or to pass it further.

---

While I agree with the notion of building a better testing API, this PR addresses an existing API introduced in another PR. Specifically, `$validateUsing` is `NULL` by default and the change here just follows that up to allow re-setting the value from the setter.

Since _this current_ test API is the only test API we, the consumer project authors, currently have, the proposed change to re-order the `match()` would allow us to use it in our tests without halting and waiting for input.

PR #193, opened 2025-06-11
Add Href prompts

luanfreitasdev:
Sometimes we want to show a clickable link to the user, so this PR adds the "href" functionality, which is supported by most current terminals on Linux and macOS.

```php
href(
    message: 'Visit Laravel documentation:',
    path: 'https://laravel.com/docs',
    tooltip: 'Click here'
);
```

image


AlexSkrypnyk:
@weitzman 
If prompts supports colours (which is extended terminal capability), why it should not support links?

luanfreitasdev:
> @luanfreitasdev I've read the code more carefully. Looks like the link is its own element. Is it possible to implement it as a formatter which could be used within _existing_ textual parts of any prompt component? E.g., in hints or labels

I understand. Any ideas for an API with a display sample?

luanfreitasdev:
Updated:

image

PR #198, opened 2025-07-21
FIX: update phpstan to latest and remove ignored file

BinaryKitten:
## Details

In previous [PR](https://github.com/laravel/prompts/pull/197) the Progress prompts file was added to alleviate that PHPStan was getting into an infinite loop and crashing - preventing CI from completing the static analysis.

After this was merged, I went to PHPStan and filed the issue with details from the Progress file. - see issue [PHPStan - Path in match causes infinite loop ( iterator|int type definition )](https://github.com/phpstan/phpstan/issues/13218 
)

This issue has now been fixed in PHPStan with new releases 1.12.28 and 2.1.15. 

This PR updates PHPStan to minimum supported requirement of 1.12.28 and removes the Progress prompt file from being ignored by PHPStan.

### DIFF
--- a/src/Progress.php
+++ b/src/Progress.php
@@ -34,10 +34,12 @@ class Progress extends Prompt
      */
     public function __construct(public string $label, public iterable|int $steps, public string $hint = '')
     {
-        $this->total = match (true) { // @phpstan-ignore assign.propertyType
+        /** @phpstan-ignore assign.propertyType (PHPStan doesn't parse that we convert from iterable to int in the below match) */
+        $this->total = match (true) {
             is_int($this->steps) => $this->steps,
             is_countable($this->steps) => count($this->steps),
             is_iterable($this->steps) => iterator_count($this->steps),
+            /** @phpstan-ignore match.unreachable (Technically we shouldn't be able to reach the default as should be int|countable|iterable ) */
             default => throw new InvalidArgumentException('Unable to count steps.'),
         };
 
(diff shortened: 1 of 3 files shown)


src/Progress.php
```diff
@@ -34,10 +34,12 @@ class Progress extends Prompt
      */
     public function __construct(public string $label, public iterable|int $steps, public string $hint = '')
     {
-        $this->total = match (true) { // @phpstan-ignore assign.propertyType
+        /** @phpstan-ignore assign.propertyType (PHPStan doesn't parse that we convert from iterable to int in the below match) */
+        $this->total = match (true) {
             is_int($this->steps) => $this->steps,
             is_countable($this->steps) => count($this->steps),
             is_iterable($this->steps) => iterator_count($this->steps),
+            /** @phpstan-ignore match.unreachable (Technically we shouldn't be able to reach the default as should be int|countable|iterable ) */
```
BinaryKitten, line 42: added ignore here with explanation

BinaryKitten:
the need for the PHPStan ignores is removed if we refactor the constructor from a match call to a series of if calls there's no need to ignore the errors as they don't exist. 
I opted to not perform the refactor as this felt larger modification than the expected update for PHPStan

```php
  public function __construct(public string $label, public iterable|int $steps, public string $hint = '')
    {
        if (is_int($this->steps)) {
            $this->total = $this->steps;
        }

        if (is_countable($this->steps)) {
            $this->total = count($this->steps);
        }

        if (is_iterable($this->steps)) {
            $this->total = iterator_count($this->steps);
        }

        if ($this->total === 0) {
            throw new InvalidArgumentException('Progress bar must have at least one item.');
        }
    }
```

PR #212, opened 2026-01-19
Enhance SearchPrompt to support array, Collection, and Closure options

sakshamgorey:
## Why

`SearchPrompt` is essentially a searchable version of `SelectPrompt`, but they don't accept the same input types. `SelectPrompt` accepts static arrays while `SearchPrompt` requires a `Closure`, even for simple static lists.

This creates an inconsistency and makes switching from `select()` to `search()` unnecessarily complex.

This PR brings `SearchPrompt` to parity with `SelectPrompt` (and `SuggestPrompt`) by accepting `array|Collection` directly for static data.

## Impact

**Before:**
```php
search(
    'Select a table',
    function (string $value) use ($tables) {
        return array_filter($tables, function ($item) use ($value) {
            return str_contains(strtolower($item), strtolower($value));
        });
    }
);
```

**After:**
```php
search('Select a table', $tables);
```

## Why Support Arrays?

1. **Consistency with other prompts:** `select()`, `multiselect()`, and `suggest()` all accept arrays for static data
2. **Simpler API:** For the common case of filtering a static list, developers shouldn't need to implement the filtering logic
3. **Backward compatible:** Closures continue to work for dynamic data (databases, APIs) or custom filtering logic

## Implementation Details

**Filtering Behavior:**  
When an array/collection is provided, filtering uses case-insensitive `str_contains()` matching.

**Return Values:**
- **List arrays** (`['A', 'B']`): Returns the selected value (e.g. `'B'`)
- **Associative arrays** (`['a' => 'A']`): Returns the selected key (e.g. `'a'`)

**When to use arrays vs closures:**
- ✅ Use arrays: Static lists where simple substring matching is sufficient
- ✅ Use closures: Database queries, API calls, or custom filtering logic (fuzzy search, starts_with, etc.)

> **Note:** This change does not replace select().
select() remains ideal for small, scannable lists, while search() is better for large or growing lists.

## Follow-Up Work

For full consistency, `multisearch()` should also accept static arrays. I'll submit that as a separate PR to keep this focused and easier to review.

## Testing

- ✅ List arrays
- ✅ Associative arrays  
- ✅ Collections
- ✅ Backward compatibility with closures

### DIFF
--- a/tests/Feature/SearchPromptTest.php
+++ b/tests/Feature/SearchPromptTest.php
@@ -248,7 +248,7 @@ it('accepts a Collection', function () {
     );
 
     expect($result)->toBe('Blue');
-});
+})->skip(! depends_on_collection());
 
 it('accepts a Collection with keys', function () {
     Prompt::fake(['u', 'e', Key::DOWN, Key::ENTER]);
@@ -263,7 +263,7 @@ it('accepts a Collection with keys', function () {
     );
 
     expect($result)->toBe('blue');
-});
+})->skip(! depends_on_collection());
 
 it('shows all options when no search input with static array', function () {
     Prompt::fake([Key::DOWN,Key::DOWN, Key::ENTER]);


sakshamgorey:
Thanks for reviewing this @taylorotwell, I completely understand the maintenance considerations.

Before closing, I wanted to share the real-world case that motivated this PR:

**Laravel's own `db:table` command** currently uses `select()` for table selection:

```php
$tableName = $this->argument('table') ?: select(
    'Which table would you like to inspect?',
    array_keys($tables)
);
```

For projects with 50+ tables, `search()` would provide better UX, but the current API makes this bit lenthy:

```php
// Current: Manual filtering logic required
search(
    'Which table would you like to inspect?',
    function (string $value) use ($tables) {
        return array_filter(
            array_keys($tables),
            fn($table) => str_contains(strtolower($table), strtolower($value))
        );
    }
);
```

With this PR, it would be as simple as swapping `select()` for `search()`:

```php
search('Which table would you like to inspect?', array_keys($tables))
```

**The filtering pattern already exists** in `select()`, `multiselect()`, and `suggest()` — this PR just extends it to `search()` for API consistency.

PR #213, opened 2026-01-21
Fix TTY handling for prompts using tmux and host php from nix

panakour:
Fixes `stty: 'standard input': unable to perform all requested operations` error when running interactive prompts inside terminal multiplexers like tmux.

Without this change if you are under a tmux and try to use a command with prompt e.g
```
❱ php artisan db:table

 ┌ Which table would you like to inspect? ──────────────────────┐
 │ public.cache_locks                                           │
 └──────────────────────────────────────────────────────────────┘


   RuntimeException

  stty: 'standard input': unable to perform all requested operations

  at laravel-prompts/src/Terminal.php:117
    113▕         $stderr = stream_get_contents($pipes[2]);
    114▕         $code = proc_close($process);
    115▕
    116▕         if ($code !== 0 || $stdout === false) {
  ➜ 117▕             throw new RuntimeException(trim($stderr ?: "Unknown error (code: $code)"), $code);
    118▕         }
    119▕
    120▕         return $stdout;
    121▕     }

  1   laravel-prompts/src/Terminal.php:55
      Laravel\Prompts\Terminal::exec("stty 6902:3:4b00:5cb:4:ff:ff:7f:17:15:12:ff:3:1c:1a:19:11:13:16:f:1:0:14:ff
")

  2   laravel-prompts/src/Prompt.php:446
      Laravel\Prompts\Terminal::restoreTty()
```

## Problem
When Laravel Prompts runs inside tmux or screen, the `stty` commands executed via `proc_open()` fail because the inherited stdin doesn't properly reference the controlling terminal through the PTY layer.
## Solution
Explicitly redirect stdin from `/dev/tty` when executing stty commands. The `/dev/tty` device always refers to the controlling terminal, bypassing any PTY indirection.

### DIFF
--- a/src/Terminal.php
+++ b/src/Terminal.php
@@ -97,7 +97,7 @@ class Terminal
      */
     protected function exec(string $command): string
     {
-        $process = proc_open($command, [
+        $process = proc_open($command.' < /dev/tty', [
             1 => ['pipe', 'w'],
             2 => ['pipe', 'w'],
         ], $pipes);


panakour:
Hey @jessarcher, thanks for looking into this!
Environment:
- macOS 26.2
- tmux 3.6a

Glad I’m not the only one experiencing this. I’ll dig into it a bit further to see whether it’s related to my tmux configuration and will let you know.

panakour:
Hey @jessarcher, I've done investigation and here's what I found:

You can see my tmux config here:
https://github.com/panakour/dotfiles/blob/master/modules/shared/tmux/tmux.conf 

The error is caused by this binding:`bind-key p run-shell "tmux-sessionizer"` 

This uses fzf-tmux -p to open a popup picker.

if you want to see the script its here: https://github.com/panakour/dotfiles/blob/master/bins/bin/tmux-sessionizer.

What I tried to fix on my side:
- stty sane after fzf-tmux - doesn't help (runs in wrong PTY context)
- display-popup -E instead of run-shell - doesn't work with fzf-tmux -p

I believe the fix adds value as a defensive measure for real-world tmux setups. But I understand if it's too niche - happy to close if you prefer.

panakour:
@jessarcher   **New finding:** fails with PHP (tested 8.5, 8.4) installed natively on macOS via Nix, but works fine
  inside Docker.

PR #229, opened 2026-03-18
Render spinner statically when --no-ansi flag is used

### LINKED ISSUE #201: Spinner animation still happens with `--no-ansi` flag
### Laravel Prompts Version

0.3.6

### Laravel Version

11.44.2

### PHP Version

8.4.12

macOS 15.6

### Terminal Application

VSCode

### Description

Spinner animation still happens with `--no-ansi` flag.

Example: This results in the animation spamming the console output.

```
⠋ Installing dependencies...
⠙ Installing dependencies...
⠹ Installing dependencies...
...and so on
```

It would be reasonable to assume `--no-ansi` would show static text and/or a static replacement icon.

### Steps To Reproduce

Standard use of `spin($callback, $message);`.

You could reproduce this in GitHub Actions or if the command is run in a Laravel Forge deployment.

- `--ansi` or no flag provided: Users would expect the animation to occur. You would expect it to print as multiple lines in a CI/CD pipeline _(animated)_
- `--no-ansi`: Users would expect the spinner to be static text _(not animated)_

JoshSalway:
## Summary
- Fixes spinner animation spamming console output when `--no-ansi` flag is passed
- Checks `isDecorated()` on the output instance and falls back to static rendering when ANSI output is disabled
- Affects CI/CD environments (GitHub Actions, Forge deployments) where `--no-ansi` is commonly used

## Test plan
- [ ] Run a command using `spin()` with `--no-ansi` flag and verify static output (no animation frames)
- [ ] Run the same command without the flag and verify animation still works normally
- [ ] Verify behavior in a CI/CD pipeline where ANSI is disabled

Fixes #201

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Spinner.php
+++ b/src/Spinner.php
@@ -47,7 +47,7 @@ class Spinner extends Prompt
     {
         $this->capturePreviousNewLines();
 
-        if (! function_exists('pcntl_fork')) {
+        if (! function_exists('pcntl_fork') || ! static::output()->isDecorated()) {
             return $this->renderStatically($callback);
         }
 


JoshSalway:
Moving to fork.

PR #234, opened 2026-04-17
Add skipUsing argument to short-circuit interactive prompts

WendellAdriel:
Interactive prompts currently force callers to wrap every call in an if/else when a command flag should pre-answer the prompt. This adds a `skipUsing` argument to every prompt helper, prompt class, and `FormBuilder` step — passing a value other than `null` or `false` (or a closure that returns one) short-circuits the prompt and returns the value after running it through the existing `transform` + `validate` pipeline. Both `null` and `false` mean "not provided", so unset options and absent boolean flags fall through to the normal interactive prompt; a negative confirm answer can be pre-supplied with a boolean-like string such as `"no"`, `"false"`, or `"0"`.

Pre-supplied values are validated like interactive input, plus a few guarantees interactive use already provides:

- `ConfirmPrompt` coerces boolean-style strings (`"1"`, `"true"`, `"yes"`, `"off"`, …) and rejects anything unparseable instead of guessing.
- `NumberPrompt` coerces numeric strings to integers and runs the usual number/min/max validation.
- `SelectPrompt`, `MultiSelectPrompt`, and `DataTablePrompt` reject values that are not in their options, so skipping can never return a value interactive use could not produce.
- `MultiSelectPrompt` and `MultiSearchPrompt` reject non-array values.

Failures throw a new `SkippedValueValidationException` (extending `NonInteractiveValidationException`, so existing catches keep working) with the prompt label and error, so invalid CLI input fails loudly instead of silently slipping through. Since `search` and `multisearch` options are closures, skipped values there can't be checked against the options — pass `validate` when that matters. A skipped prompt writes no output, matching the existing non-interactive behavior. Form steps that skip are ignored when reverting with CTRL+U — closure resolvers are re-evaluated to decide — so users revert past pre-answered steps to the previous interactive prompt. `pause` is intentionally excluded.

## Example

```php
use function Laravel\Prompts\confirm;

// Wire an Artisan option straight into the prompt — no if/else needed.
$name = text(
    label: 'What is your name?',
    required: true,
    skipUsing: $this->option('name'), // null → prompt; 'Taylor' → returns 'Taylor'
);

// Boolean flags work directly — false means "not provided", like null.
$confirmed = confirm(
    label: 'Continue?',
    skipUsing: $this->option('yes'), // --yes → true; absent → prompt
);
```

### DIFF
--- a/src/AutoCompletePrompt.php
+++ b/src/AutoCompletePrompt.php
@@ -39,7 +39,10 @@ class AutoCompletePrompt extends Prompt
         public mixed $validate = null,
         public string $hint = '',
         public ?Closure $transform = null,
+        mixed $skipWhen = null,
     ) {
+        parent::__construct($skipWhen);
+
         $this->options = $options instanceof Collection ? $options->all() : $options;
 
         $this->on('key', function ($key) {
--- a/src/AutoCompletePrompt.php
+++ b/src/AutoCompletePrompt.php
@@ -39,9 +39,9 @@ class AutoCompletePrompt extends Prompt
         public mixed $validate = null,
         public string $hint = '',
         public ?Closure $transform = null,
-        mixed $skipWhen = null,
+        mixed $skipUsing = null,
     ) {
-        parent::__construct($skipWhen);
+        parent::__construct($skipUsing);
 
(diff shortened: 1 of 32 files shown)


taylorotwell:

PR #236, opened 2026-05-13
Add default values support to multisearch prompt

saeedhosseiinii:
## Summary

This adds support for default selected values to the `multisearch` prompt.

The new `default` argument is appended to the existing function and form builder signatures so existing positional calls remain backward compatible. When provided, default values are initialized as selected before the prompt is rendered.

## Notes

- Supports both array and collection defaults.
- Keeps existing `multisearch()` positional argument order unchanged.
- Matches the existing `multiselect()` default value behavior more closely.


taylorotwell:

PR #242, opened 2026-05-24
Support terminal extension backend

prateekbhujel:
This lets Prompts use the optional `terminal` PHP extension as its terminal backend when it is installed.

Context: I have been working on a small PHP extension for native terminal helpers here:

https://github.com/prateekbhujel/php-terminal

The reason for this PR is Windows support. Prompts currently rejects Windows before entering the interactive loop because the current backend depends on `stty` / Unix-style terminal input. The extension exposes the pieces Prompts needs through PHP functions instead: TTY checks, raw mode, safe restore, single-key reads, and terminal size.

This PR keeps the extension completely optional. If it is not installed, Prompts keeps the existing behavior. If it is installed, Prompts can use the extension backend and map its named keys back to the existing `Key::*` sequences, so the prompt code itself does not need to learn a second key format.

The extension has a `v0.1.0` alpha release with Windows builds here:

https://github.com/prateekbhujel/php-terminal/releases/tag/v0.1.0

Tests run:
- `vendor/bin/pest tests/Feature/TerminalTest.php`
- `vendor/bin/pest`
- `vendor/bin/phpstan analyse --memory-limit=1G`
- `php -d extension=/tmp/php-terminal-build-x86/modules/terminal.so vendor/bin/pest`

### DIFF
--- a/src/Prompt.php
+++ b/src/Prompt.php
@@ -108,7 +108,7 @@ abstract class Prompt
                 return $this->fallback();
             }
 
-            static::$interactive ??= stream_isatty(STDIN);
+            static::$interactive ??= static::terminal()->interactive();
 
             if (! static::$interactive) {
                 return $this->default();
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #248, opened 2026-06-22
Print completion line after task finishes

lazerg:
Fixes #246

After `task()` finishes, the label and spinner disappear entirely with nothing left on screen. Added a `printCompletionLine()` method that writes `✓ {label}` after the spinner block is erased, called from both the `resetTerminal()` and `renderStatically()` paths.


joetannenbaum:
Thank you for this! Decided to go a different way with it as it's a very similar component to `spin` and the default behavior there is to disappear as well. So if you pass the `keepSummary` argument and there aren't any "stable" messages, it will display the completion message that you added 👍

PR #249, opened 2026-06-22
Use static rendering for Spinner when output is not decorated

### LINKED ISSUE #201: Spinner animation still happens with `--no-ansi` flag

lazerg:
When the `--no-ansi` flag is passed, the output is not decorated and cursor-movement escape sequences have no effect. Previously the spinner would still fork a child process that wrote animation frames in a loop, flooding the console with lines like `⠋ Running...`, `⠙ Running...`, and so on indefinitely. Now, when `isDecorated()` returns false, the spinner falls back to `renderStatically()` — the same path taken when `pcntl_fork` is unavailable — so only a single static frame is shown.

Fixes #201.

### DIFF
--- a/src/Spinner.php
+++ b/src/Spinner.php
@@ -47,7 +47,7 @@ class Spinner extends Prompt
     {
         $this->capturePreviousNewLines();
 
-        if (! (function_exists('pcntl_fork') && function_exists('posix_kill'))) {
+        if (! static::output()->isDecorated() || ! (function_exists('pcntl_fork') && function_exists('posix_kill'))) {
             return $this->renderStatically($callback);
         }
 
--- a/src/Task.php
+++ b/src/Task.php
@@ -109,7 +109,7 @@ class Task extends Prompt
 
         $this->capturePreviousNewLines();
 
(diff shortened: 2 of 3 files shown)


joetannenbaum:
Thank you!

PR #272, opened 2026-09-16
Properly handle CSI escape sequences in InteractsWithStrings

prateekbhujel:
Currently stripEscapeSequences uses a regex that assumes all escape sequences end in m. When text includes other standard control sequences such as line clears or cursor movement like \e[2K or \e[1A, the regex matches across following text until the first letter m in ordinary words, deleting user text in between.

Similarly, parseAnsiText only terminates when matching sequences ending in m, causing other control sequences to consume the rest of the text to the end of the string.

This updates stripEscapeSequences to match standard CSI patterns and ensures parseAnsiText terminates control sequences properly without dropping subsequent text.


taylorotwell:

